PR #6, opened 2015-09-18
main: emit warnings and fail when input contains unrecognized keywords

vcaputo:
We want to fail when the input is valid YAML but refers to keywords not
captured by the ignition config schema.

### DIFF
--- a/src/main.go
+++ b/src/main.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"reflect"
 	"strings"
 
 	"github.com/coreos/fuze/third_party/github.com/coreos/ignition/config"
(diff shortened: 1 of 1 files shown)


vcaputo:
I'm new to the reflect side of golang, so this might be totally batshit crazy.

src/main.go
```diff
@@ -31,6 +32,45 @@ func stderr(f string, a ...interface{}) {
...
+			for ; i < refType.NumField(); i++ {
+				sf = refType.Field(i)
+				tv := sf.Tag.Get("yaml")
+				if tv == key {
+					break
+				}
+			}
+
+			if i < refType.NumField() {
+				if warn := hasUnrecognizedKeys(ks[key], sf.Type); warn {
```
crawford: I think this is a little more clear when it is in the loop.

src/main.go
```diff
@@ -31,6 +32,45 @@ func stderr(f string, a ...interface{}) {
...
+					warnings = true
```
crawford: |= rather than the branch?

src/main.go
```diff
@@ -63,6 +103,17 @@ func main() {
...
 
+	var inCfg interface{}
+	if err := yaml.Unmarshal(dataIn, &inCfg); err != nil {
+		stderr("Failed to unmarshal input: %v", err)
+		os.Exit(1)
+	}
+
+	if hasUnrecognizedKeys(inCfg, reflect.TypeOf(cfg)) {
+		stderr("Unrecognized keys in input, aborting.")
+		os.Exit(1)
```
crawford: I'm not sure this should be a failure. Why not a warning?

crawford:
A few NITs but the approach looks good.

src/main.go
vcaputo: |= for a bool in golang?

src/main.go
vcaputo: I'm not sure either, but it seemed appropriate to fail since unrecognized fields aren't going to be preserved in the json output.  If they're in the ignition yaml input presumably it's desired in the output, failing seems appropriate.  Maybe add a flag to ignore warnings?

crawford: Actually, I think the failure is correct.

src/main.go
crawford: Oh for fuck's sake. Nevermind.

src/main.go
```diff
@@ -31,6 +32,44 @@ func stderr(f string, a ...interface{}) {
...
+			i := 0
+			sf := reflect.StructField{}
+			for ; i < refType.NumField(); i++ {
+				sf = refType.Field(i)
+				tv := sf.Tag.Get("yaml")
+				if tv == key {
+					if warn := hasUnrecognizedKeys(ks[key], sf.Type); warn {
+						warnings = true
+					}
+					break
```
crawford: If you continue to the outer loop, you can narrow the scope of `i` and remove the condition on the warning.

src/main.go
```diff
@@ -31,6 +32,44 @@ func stderr(f string, a ...interface{}) {
 	fmt.Fprintln(os.Stderr, strings.TrimSuffix(out, "\n"))
 }
 
+// hasUnrecognizedKeys finds unrecognized keys and warns about them on stderr.
+// returns false when no unrecognized keys were found, true otherwise.
+func hasUnrecognizedKeys(inCfg interface{}, refType reflect.Type) (warnings bool) {
+
```
crawford: Extra space. I don't know why go fmt doesn't fix that.

src/main.go
vcaputo: good point, labelled continues haven't set in yet.

PR #7, opened 2015-09-22
main: add test for hasUnrecognizedKeys()

vcaputo:


src/main_test.go
```diff
@@ -15,9 +15,44 @@
...
+		{
+			in:    "ignition_version: 1",
+			unrec: false,
+		},
+		{
+			in:    "ignition_version: 1\npasswd:\n users:\n  - name: foobar\n",
+			unrec: false,
+		},
+		{
+			in:    "foo: bar",
```
jonboulle: IMO might be nice to return a more specific error if `ignition_version` is missing since in principle that field is what defines it as an ignition configuration and not just some random json. Just a thought.

jonboulle:
lgtm

src/main_test.go
vcaputo: makes sense

PR #8, opened 2015-11-09
hasUnrecognizedKeys failed to dereference pointer

dgonyeo:
In some cases hasUnrecognizedKeys can apparently be handed a pointer to
a struct, as opposed to just a struct. This commit checks if refType is
a pointer, and dereferences it if it is.

### DIFF
--- a/src/main.go
+++ b/src/main.go
@@ -35,6 +35,9 @@ func stderr(f string, a ...interface{}) {
 // hasUnrecognizedKeys finds unrecognized keys and warns about them on stderr.
 // returns false when no unrecognized keys were found, true otherwise.
 func hasUnrecognizedKeys(inCfg interface{}, refType reflect.Type) (warnings bool) {
+	if refType.Kind() == reflect.Ptr {
+		refType = refType.Elem()
+	}
 	switch inCfg.(type) {
 	case map[interface{}]interface{}:
 		ks := inCfg.(map[interface{}]interface{})


src/main.go
```diff
@@ -35,6 +35,9 @@ func stderr(f string, a ...interface{}) {
 // hasUnrecognizedKeys finds unrecognized keys and warns about them on stderr.
 // returns false when no unrecognized keys were found, true otherwise.
 func hasUnrecognizedKeys(inCfg interface{}, refType reflect.Type) (warnings bool) {
+	if refType.Kind().String() == "ptr" {
```
crawford: I think you want [Indirect](https://golang.org/pkg/reflect/#Indirect).

dgonyeo: Yup, I do. The reflect package has a lot of things in it.

dgonyeo: Err, maybe not. `refType` is a `reflect.Type`, but `reflect.Indirect()` takes a `reflect.Value`, and the `reflect.Type` interface doesn't have an `Indirect` function.

crawford: :/ OK, how about comparing against reflect.Ptr instead of the string then?

dgonyeo: Missed that, yeah that's a lot cleaner. Done.

crawford:
LGTM. @vcaputo?

vcaputo:
Neat, lgtm!

PR #14, opened 2016-05-27
config: add initial config and validation

crawford:


crawford:
/cc @dghubble 

This doesn't work yet, but I wanted feedback on the config syntax.

config/types.go
```diff
@@ -0,0 +1,126 @@
...
+	} `yaml:"passwd"`
+}
+
+type ConfigReference struct {
+	Source       string       `yaml:"source"`
+	Verification Verification `yaml:"verification"`
+}
+
+type Verification struct {
+	Hash *string `yaml:"hash"`
```
dghubble: This would default to sha512 instead of requiring the Hash `Function` and `Sum` be specified each time?

crawford: It is specified as "sha512-5673..." in the Ignition config. I just used the exact same here.

config/types.go
```diff
@@ -0,0 +1,126 @@
...
+		} `yaml:"config"`
+	} `yaml:"ignition"`
+	Storage struct {
+		Disks []struct {
+			Device     string `yaml:"device"`
+			WipeTable  bool   `yaml:"wipe_table"`
+			Partitions []struct {
+				Label    string `yaml:"label"`
+				Number   int    `yaml:"number"`
+				Size     string `yaml:"size"`
```
dghubble: Since this is a string, the plan is to continue supporting MiB, GiB dimensions I'd assume

crawford: Correct

config/types.go
```diff
@@ -0,0 +1,126 @@
...
+		Files []struct {
+			Filesystem string `yaml:"filesystem"`
+			Path       string `yaml:"path"`
+			Contents   struct {
+				Remote struct {
+					Url          string       `yaml:"url"`
+					Compression  string       `yaml:"compression"`
+					Verification Verification `yaml:"verification"`
+				} `yaml:"remote"`
+				Inline string `yaml:"inline"`
```
dghubble: :D

config/types.go
```diff
@@ -0,0 +1,126 @@
...
+			} `yaml:"contents"`
+			Mode int `yaml:"mode"`
```
dghubble: Before this was an alias of os.FileMode for validation and unmarshaling. Maybe we'd to the same here since authors will write the wrong mode or provide decimal vs octal.

dghubble:
`Config` appears good to me and has the needed inline feature:

```
files:
  filesystem: "root"
  path: /etc/kuberenetes/manifests/component.yaml
  contents:
    inline: |
      file
      contents
      here
```

config/types.go
crawford: The general idea is to use dumb types in this config, do a 1-1 translation to Ignition, and then use Ignition's built-in verification.

dghubble: Ok, disregard that.

config/types.go
crawford: I've broken this out into a sum and function field.

crawford:
Needs docs. Needs Ignition validation.

config/config.go
```diff
@@ -0,0 +1,48 @@
...
+
+	"github.com/coreos/ignition/config/types"
+	"github.com/go-yaml/yaml"
+)
+
+func ParseAsNewest(data []byte) (types.Config, error) {
+	return ParseAsV2_0_0(data)
+}
+
+func ParseAsOldest(data []byte) (types.Config, error) {
```
dghubble: Consuming this lib, I can't think of a case I'd prefer `ParseAsOldest` over the specific `ParseAsV2_0_0` or whatever it is? I'd prefer obvious breakage if `ParseV2_0_0` were removed over the min version quietly bumping up.

crawford: There was a particular use case I had in mind when I wrote that, but I can't remember it now. Let's remove it.

dghubble:
`PartitionTypeGUID` doesn't seem to default the empty string value. `partition type-guid must have the form "01234567-89AB-CDEF-EDCB-A98765432101", got: ""`

config/convert.go
```diff
@@ -0,0 +1,267 @@
...
+	for _, file := range in.Storage.Files {
+		newFile := types.File{
+			Filesystem: file.Filesystem,
+			Path:       types.Path(file.Path),
+			Mode:       types.FileMode(file.Mode),
+			User:       types.FileUser{Id: file.User.Id},
+			Group:      types.FileGroup{Id: file.Group.Id},
+		}
+
+		if file.Contents.Inline != "" {
```
dghubble:
```
files:
    - path: /etc/kubernetes/empty
      filesystem: somefs
      mode: 0644
      contents:
        inline: |
```

Adding that to a fuze config winds up dropping to an emergency shell on boot because no Inline contents falls through to the URL scheme below.

createFiles: fetching file /etc/kubernetes/empty: unsupported source scheme
failed to create files: failed to resolve file /etc/kubernetes/empty

In one case, an empty file like this was used for the side-effect of ensuring the `/etc/kubernetes` directory existed. Its easily worked around by just adding some dummy file contents for now.

crawford: This is actually a bug in Ignition (https://github.com/coreos/bugs/issues/1420). I'm curious how we should handle this case. We could fix Ignition and bump the spec to 2.0.1. Then fuze could include a workaround if it sees 2.0.0.

crawford:
```yml
storage:
  files:
    - path: /etc/kubernetes/empty
      filesystem: somefs
      mode: 0644
```

results in:

```json
{
  "ignition": {
    "version": "2.0.0",
    "config": {}
  },
  "storage": {
    "files": [
      {
        "filesystem": "somefs",
        "path": "/etc/kubernetes/empty",
        "contents": {
          "source": "data:,",
          "verification": {}
        },
        "mode": 420,
        "user": {},
        "group": {}
      }
    ]
  },
  "systemd": {},
  "networkd": {},
  "passwd": {}
}
```

config/convert.go
dghubble: By "if it sees 2.0.0", you mean if ParseAsV2_0_0 is called vs ParseAsV2_0_1?

crawford: Correct.

dghubble: Sounds great, kinda handy to have this fuze intermediate layer.

dghubble:
This works well for me and I'm happy to work around the empty file issue in the fuze configs we use. Even cooler if it is handled. LGTM.

crawford:
It will be handled for you.

PR #15, opened 2016-08-30
Replace remote file "source" key with "url" 

bzub:
There's a typo in the specification.  The other option is to change the key to "source" in the code so it matches that of ignition append, and ignition JSON.


crawford:
LGTM. Thank you.

crawford:
Can you reformat the commit message to comply with https://github.com/coreos/fuze/blob/master/CONTRIBUTING.md#format-of-the-commit-message?

PR #17, opened 2016-09-07
Update Ignition to v0.10.1, handle validation changes

dghubble:
Fuze's vendored Ignition is preventing me from updating coreos-baremetal Ignition beyond v0.7.0. Ignition has a breaking change from AssetValid to a Report based model, which hasn't been applied here in Fuze. https://github.com/coreos/ignition/commit/cbf3871bbe9d8dbff3b525f8b783532c9ba2b997

Please lemme know how you want to proceed or if you want a different approach? This PR will fail with the current tests.


crawford:
I would just propagate the report through and allow main to pretty print it for now (it won't be extremely accurate since it doesn't have a JSON source to reference). @ajeddeloh is working on improving the reporting for Fuze.

dghubble:
I'll wait on @ajeddeloh's reporting updates then.

PR #18, opened 2016-09-16
WIP: config: add etcd and flannel configuration

ajeddeloh:
Adds options to allow directly configuring etcd, etcd2, and flannel.


config/convert.go
```diff
@@ -222,6 +224,64 @@ func ConvertAs2_0_0(in Config) (types.Config, error) {
 		})
 	}
 
+	// etcd and etcd2 get drop in files with the configuration
+	if !isZero(in.Etcd) {
+		out.Systemd.Units = append(out.Systemd.Units, types.SystemdUnit{
+			Name:   "etcd.service",
+			Enable: true,
+			DropIns: []types.SystemdUnitDropIn{{
+				Name:     "20-ignition-etcd.conf",
```
crawford: Let's call this "20-fuze-etcd.conf" instead.

config/convert.go
```diff
@@ -222,6 +224,64 @@ func ConvertAs2_0_0(in Config) (types.Config, error) {
...
+			}},
+		})
+	}
+
+	if !isZero(in.Etcd2) {
+		out.Systemd.Units = append(out.Systemd.Units, types.SystemdUnit{
+			Name:   "etcd2.service",
+			Enable: true,
+			DropIns: []types.SystemdUnitDropIn{{
+				Name:     "20-ignition-etcd2.conf",
```
crawford: Same here.

config/convert.go
```diff
@@ -222,6 +224,64 @@ func ConvertAs2_0_0(in Config) (types.Config, error) {
...
+
+	// flannel gets a /run/flannel/options.env" file with the configuration
+	if !isZero(in.Flannel) {
+		envVars := strings.Join(getEnvVars(in.Flannel), "\n")
+		if envVars != "" {
+			// Since the files go in a tmpfs, we need to tell ignition about the /run filesystem
+			// XXX How do we ensure __runTmpfs doesn't interfere with user supplied filesystem names?
+			runTmpfs := types.Filesystem{
+				Name: "__runTmpfs",
+				Path: func(tmp types.Path) *types.Path { return &tmp }(types.Path("/run")),
```
crawford: I think you can omit the cast to `types.Path`.

config/convert.go
```diff
@@ -222,6 +224,64 @@ func ConvertAs2_0_0(in Config) (types.Config, error) {
...
+			}},
+		})
+	}
+
+	// flannel gets a /run/flannel/options.env" file with the configuration
+	if !isZero(in.Flannel) {
+		envVars := strings.Join(getEnvVars(in.Flannel), "\n")
+		if envVars != "" {
+			// Since the files go in a tmpfs, we need to tell ignition about the /run filesystem
+			// XXX How do we ensure __runTmpfs doesn't interfere with user supplied filesystem names?
```
crawford: Well, you do know all of the filesystems provided by the user. You could just double check that the name doesn't collide.

config/types.go
```diff
@@ -114,6 +114,111 @@ type Config struct {
 			System       bool   `yaml:"system"`
 		} `yaml:"groups"`
 	} `yaml:"passwd"`
+	Etcd struct {
```
crawford: We are going to drop etcd0, so you can probably just skip this section.

config/types.go
```diff
@@ -114,6 +114,111 @@ type Config struct {
...
+		Proxy                    string  `yaml:"proxy"                         env:"ETCD_PROXY"                       deprecated:"etcd2 options no longer work for etcd"`
+		RetryInterval            float64 `yaml:"retry_interval"                env:"ETCD_RETRY_INTERVAL"`
+		Snapshot                 bool    `yaml:"snapshot"                      env:"ETCD_SNAPSHOT"`
+		SnapshotCount            int     `yaml:"snapshot_count"                env:"ETCD_SNAPSHOTCOUNT"`
+		StrTrace                 string  `yaml:"trace"                         env:"ETCD_TRACE"`
+		Verbose                  bool    `yaml:"verbose"                       env:"ETCD_VERBOSE"`
+		VeryVerbose              bool    `yaml:"very_verbose"                  env:"ETCD_VERY_VERBOSE"`
+		VeryVeryVerbose          bool    `yaml:"very_very_verbose"             env:"ETCD_VERY_VERY_VERBOSE"`
+	} `yaml:"etcd"`
+	Etcd2 struct {
```
crawford: We are planning on shipping etcd in a container instead of baking it into the OS. Since there will be the option of many different versions being run, the set of environment variables will need to change based on the version being run. I wonder if there is anything we can do to help with that.

ajeddeloh: We'll need some mapping of etcd version to valid options. Once I port over the ignition Validate() code, then we could have the etcd struct be a version field and list of KV pairs. You could then use Validate() to verify they match the version correctly, which would make interface the same for all etcd versions.

crawford: That seems reasonable. I wonder if there is a way we can easily discover that information from etcd itself, instead of having to maintain both etcd and these PV pairs. /cc @heyitsanthony

ajeddeloh:
DO NOT MERGE
Tests not done

ajeddeloh:
Updated to use new ignition (v2_0 types) with validation fix
Made etcd/flannel set env's for the etcd-member.service and flanneld.service unit files.
Also updated the supported versions of flannel and etcd to be current with the latest releases

config/types/etcd.go
```diff
@@ -0,0 +1,250 @@
...
+		"v2.3.6",
+		"v2.3.5",
+		"v2.3.4",
+		"v2.3.3",
+		"v2.3.2",
+		"v2.3.1",
+		"v2.3.0",
+	}
+	for _, tag := range validTags {
+		if string(e) == tag {
```
dgonyeo: Maybe it would be worth running these through the semver library to just check major versions? Otherwise we'll need a new clct release whenever there's a new etcd release.

config/types/etcd.go
```diff
@@ -0,0 +1,250 @@
...
+		}
+		etcd.Options = o
+	} else if strings.HasPrefix(string(etcd.Version), "v3.1") {
+		o := Etcd3_1{}
+		if err := unmarshal(&o); err != nil {
+			return err
+		}
+		etcd.Options = o
+	}
+	return nil
```
dgonyeo: Will this function result in not returning an error if the etcd prefix doesn't match any of the expected values?

config/types/flannel.go
```diff
@@ -0,0 +1,115 @@
...
+		"v0.5.6",
+		"v0.5.5",
+		"v0.5.4",
+		"v0.5.3",
+		"v0.5.2",
+		"v0.5.1",
+		"v0.5.0",
+	}
+	for _, tag := range validTags {
+		if string(fv) == tag {
```
dgonyeo: Same comment as above where maybe we don't want to hard code the versions

config/types/flannel.go
```diff
@@ -0,0 +1,115 @@
...
+		}
+		flannel.Options = o
+	} else if strings.HasPrefix(string(flannel.Version), "v0.5") {
+		o := Flannel0_5{}
+		if err := unmarshal(&o); err != nil {
+			return err
+		}
+		flannel.Options = o
+	}
+	return nil
```
dgonyeo: Same comment as above with the prefixes not matching

config/types/etcd.go
ajeddeloh: Semvar would probably be a good idea. I think it's a good idea to keep up to date etcd/flannel patch releases and validate them. We could error on non-existent/too old versions and warn on too new versions (e.g. someone using an older version of clct than the latest etcd).

dgonyeo: I think we should just warn if the first two numbers are above our expected bounds. If there's a 3.1.x patch released it shouldn't impact anything enough for clct to care imo

config/types/etcd.go
ajeddeloh: It will. The question is does it make sense to fail here or fail in Validate(). If you fail in Validate() you can do things like allow newer minor versions but just not support new fields (and unmarshal to the newest minor version for that major version) and throw a warning. It also then allows multiple errors to be returned at once.

Regardless, it shouldn't return nil. If we want Validate() to catch it, it needs to return an Etcd struct with just Version set, but Options as ni.

dgonyeo: Failing in `Validate` is fine as long as it fails somewhere. If this isn't going to look like it returning an error, I think I'd like some comment here explaining why.

config/types/etcd.go
ajeddeloh: Ok, I'm cool with dropping patch, since they shouldn't impact much. On too old though, I think we should error, since we wont have a EtcdX_Y struct to unmarshal to and all their options will be dropped. On too new we should warn and state any options introduced since the latest known minor will be ignored.

dgonyeo: SGTM

dgonyeo:
Talked to @ajeddeloh, it looks like the next feature I want to add requires the generated etcd service to put user-entered arguments in the command line, instead of environment variables. Andrew is going to make the changes here, instead of me altering how this works in my future PR.

dgonyeo:
oh nevermind, I'll be changing the etcd arguments to be in the ExecStart part of the unit in my PR.

config/types/etcd.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+func (e EtcdVersion) Validate() report.Report {
+	v := semver.Version(e)
+	switch {
+	case v.LessThan(OldestEtcd):
+		return report.ReportFromError(EtcdVersionTooOld, report.EntryError)
+	case v.Major == 2 && v.Minor > 3:
+	case v.Major == 3 && v.Minor > 1:
```
dgonyeo: golang case statements don't auto-fallthrough, you'll need to use `fallthrough` explicitly: https://golang.org/ref/spec#Fallthrough_statements

dgonyeo:
@crawford want to give this another pass? I'm waiting on it for some other stuff.

crawford:
Awesome work!

dgonyeo:
```
derek@proton 148 ~/go/src/github.com/coreos/container-linux-config-transpiler> ./bin/ct <test2.yaml --pretty
error: strconv.ParseInt: parsing "v3": invalid syntax
Failed to parse config
derek@proton 1 ~/go/src/github.com/coreos/container-linux-config-transpiler> cat test2.yaml 
etcd:
  version: v3.0.15
  advertise_client_urls:       "{public.IPv4}:2379"
  initial_advertise_peer_urls: "{private.IPv4}:2380"
  listen_client_urls:          "http://0.0.0.0:2379"
  listen_peer_urls:            "http://{private.IPv4}:2380"
  initial_cluster:             "%m=http://{private.IPv4}:2380"
passwd:
  users:
    - name: core
      password_hash: "..."
```

(yes I know the templating logic won't work here)

It succeeds if the version doesn't have a `v` in there, but the v is necessary due to the [image tags on Quay](https://quay.io/repository/coreos/etcd?tab=tags) including them.

crawford:
I'd rather the version field expected the version string rather than a tag. We can derive the tag from the version.

ajeddeloh:
Rebased and fixed it so the v is prepended when the unit it generated

dgonyeo:
@crawford any final words before I merge?

crawford:
On 03/09, Sergiusz Urbaniak wrote:
> Now before digging deeper trying to find out if we have a bug in our
> Terraform setup, I wanted to make sure if we are aware of
> https://github.com/kubernetes/kubernetes/issues/40182?
> 
> Apparently upstream Docker 1.13 seems to be broken in conjunction with
> upstream k8s by changing the iptables forward policy to DROP.
> 
> Did we revert the Docker upstream commit
> https://github.com/docker/docker/pull/28257?

We are tracking the issue here:
https://github.com/coreos/bugs/issues/1842

The OS team will not promote docker 1.13.0 past Alpha. Do we need to
revert that docker commit you mentioned to unblock the OpenStack work or
are you okay waiting for the next Beta?

-Alex

crawford:
:ship: :it:

crawford:
Who is working with the Tectonic team to get this fixed in Kubernetes? Was
the Tectonic team informed?

On Thu, Mar 9, 2017 at 10:42 AM Alex Crawford <alex.crawford@coreos.com>
wrote:

On 03/09, Sergiusz Urbaniak wrote:
> Now before digging deeper trying to find out if we have a bug in our
> Terraform setup, I wanted to make sure if we are aware of
> https://github.com/kubernetes/kubernetes/issues/40182?
>
> Apparently upstream Docker 1.13 seems to be broken in conjunction with
> upstream k8s by changing the iptables forward policy to DROP.
>
> Did we revert the Docker upstream commit
> https://github.com/docker/docker/pull/28257?

We are tracking the issue here:
https://github.com/coreos/bugs/issues/1842

The OS team will not promote docker 1.13.0 past Alpha. Do we need to
revert that docker commit you mentioned to unblock the OpenStack work or
are you okay waiting for the next Beta?

-Alex

crawford:
On 03/09, Brandon Philips wrote:
> Who is working with the Tectonic team to get this fixed in Kubernetes?

No one at the moment. We have higher priority tasks eating up our time.
We are just going to keep it in Alpha until we can fix it.

> Was the Tectonic team informed?

Yes, multiple people were made aware.

-Alex

crawford:
I am fine with the next Beta as it will include Ignition 0.13 being the
only prerequisite for OpenStack.

PR #19, opened 2016-10-17
Use Ignition's validation

ajeddeloh:
Make fuze use ignition's validation.


config/vendor/github.com/coreos/ignition/config/validate/astjson/node.go
```diff
@@ -0,0 +1,59 @@
+package node
```
crawford: Can you break this into a separate commit?

config/astyaml.go
```diff
@@ -0,0 +1,76 @@
+package config
```
crawford: License header

config/config.go
```diff
@@ -27,14 +35,24 @@ func ParseAsV2_0_0(data []byte) (types.Config, error) {
...
-	if err := yaml.Unmarshal(data, &keyMap); err != nil {
-		return types.Config{}, err
-	}
+	var r report.Report
 
-	if err := assertKeysValid(keyMap, reflect.TypeOf(Config{})); err != nil {
-		return types.Config{}, err
+	nodes := yaml.UnmarshalToNode(data)
+	if nodes == nil {
+		fmt.Println("Failed to unmarshal yaml for reporting line numbers.")
```
crawford: I'm not a fan of these prints. This package is consumed by other applications. They should be able to print the way they like. Can we just return the report instead?

config/convert.go
```diff
@@ -222,8 +229,13 @@ func ConvertAs2_0_0(in Config) (types.Config, error) {
 		})
 	}
 
-	if err := out.AssertValid(); err != nil {
-		return types.Config{}, err
+	r := validate.ValidateWithoutSource(reflect.ValueOf(out))
+	if r.String() != "" {
+		fmt.Println("Ignition report:")
```
crawford: Can we return the report instead?

config/config.go
ajeddeloh: Both fuze's validation and ignition's validation with return reports. Ignition's will not have line numbers since there is no corresponding json source. Should we merge the two and return that? That would lose which source it came from (although we could iterate over the report and add that to the entry messages). 

Another possibility is to refactor so this doesn't return a ignition config structure and instead just parses the yaml and returns a the fuze config structure and fuze report. It's caller can then use that result to call ConvertAs2_0_0. Then it's up to the caller to determine how the reports should be handled.

config/types/Passwd.go
```diff
@@ -0,0 +1,47 @@
+// Copyright 2016 CoreOS, Inc.
```
crawford: lower case `passwd.go`

internal/main.go
```diff
@@ -56,9 +56,17 @@ func main() {
...
+	stderr(report.String())
+	if report.IsFatal() {
+		stderr("Failed to parse fuze config")
+		os.Exit(1)
+	}
+
+	cfg, report := config.ConvertAs2_0_0(fuzeCfg)
+	stderr(report.String())
+	if report.IsFatal() {
+		stderr("Generated ignition config was invalid.")
```
crawford: Ignition

config/types/storage.go
```diff
@@ -15,7 +15,7 @@
 package types
 
 type Storage struct {
-	Disks       []Disk       `yaml:"disk"`
```
crawford: Can you squash this change?

config/types/config.go
```diff
@@ -23,7 +23,7 @@ type Config struct {
 }
 
 type Ignition struct {
-	Config IgnitionConfig
+	Config IgnitionConfig `yaml:"config"`
```
crawford: Can you squash this change?

PR #28, opened 2017-01-23
build-docker creates a scratch Docker image

jkt628:
Usage:
docker run -i --rm coreos/ct <ignition.yaml >ignition.json

### DIFF
--- a/Dockerfile.build-exec
+++ b/Dockerfile.build-exec
@@ -1,0 +1,9 @@
+FROM golang:alpine
+ENV CGO_ENABLED=0
+WORKDIR /go/src/github.com/coreos/container-linux-config-transpiler
+COPY . .
+RUN apk update \
+    && apk add bash \
+    && ./build \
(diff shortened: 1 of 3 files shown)


dgonyeo:
My googling doesn't appear to be pulling up this answer quickly, what's the difference between `Dockerfile` and `Dockerfile.build`? One appears to rely on the binary built outside of docker, but when is one used over the other?

Aside from that, looks fine.

jkt628:
1. build-docker uses Dockerfile.build to create a container that uses the build script to create the executable.
2. then build-docker uses that executable and Dockerfile to create a tiny "from scratch" image.

dgonyeo:
I think I would prefer you to do that, otherwise new users might try to just `docker build` when they see a `Dockerfile`, and the resulting image wouldn't work since libc would be missing (assuming they've already built ct).

jkt628:
and there you go....

PR #29, opened 2017-01-26
Add version package and flag

bgilbert:
Fixes coreos/bugs#1741.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,3 +4,13 @@ The config transpiler ("ct" for short) is the utility responsible for transformi
 
 [ignition]: https://github.com/coreos/ignition
 [overview]: doc/overview.md
+
+## Building
+
+```shell
+git clone --branch v0.1.0 https://github.com/coreos/container-linux-config-transpiler
+cd container-linux-config-transpiler
+./build
+```
+
+The `ct` binary will be placed in `./bin/`.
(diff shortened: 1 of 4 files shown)


bgilbert:
Sure.  Ignition assumes that it's always built from Git, not from a tarball; is that assumption safe here?

crawford:
Assuming this will be built from a git clone is safe. We'll want to add instructions for building this to the README (git clone, git checkout tag, ./build)

README.md
```diff
@@ -4,3 +4,15 @@ The config transpiler ("ct" for short) is the utility responsible for transformi
 
 [ignition]: https://github.com/coreos/ignition
 [overview]: doc/overview.md
+
+## Building
+
+```shell
+git clone https://github.com/coreos/container-linux-config-transpiler
```
crawford: How about `git clone --branch v0.1.0 https://github.com/coreos/container-linux-config-transpiler`?

internal/main.go
```diff
@@ -50,6 +53,11 @@ func main() {
 		return
 	}
 
+	if flags.version {
+		fmt.Printf("%s\n", version.String)
```
crawford:
```go
fmt.Printf("ct %s\n", version.String)
```

bgilbert: Following the Ignition example, I put `ct` inside `version.String`.  I can move it though.

crawford: Ah, I missed that. No need to use a formatting print then.

crawford:
Looks good. Can you update the commit messages to match the rest of the history?

PR #30, opened 2017-02-23
main: initialize inFile/outFile

s-urbaniak:
Currently if one calls the transpiler with -inFile and/or -outFile
settings, an error occurs:

$ ct -in-file user_data.yaml
Failed to read: invalid argument

This is because the inFile/outFile pointers are actually not
initialized.

This fixes it.

/cc @alexsomesan
/cc @crawford

### DIFF
--- a/internal/main.go
+++ b/internal/main.go
@@ -60,17 +60,28 @@ func main() {
 
 	var inFile *os.File
 	var outFile *os.File
+	var err error
 
 	if flags.inFile == "" {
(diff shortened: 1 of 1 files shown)


alexsomesan:
👍  make more sense

crawford:
Woops, @dgonyeo sneaked in a fix (https://github.com/coreos/container-linux-config-transpiler/pull/31).

PR #33, opened 2017-02-28
main: add strict flag

crawford:
When invoked with this flag, ct will fail unless there are no errors,
warnings, deprecations, etc..


dgonyeo:
Needs rebase though

PR #35, opened 2017-03-09
intermediate change broke 'docker build'

jkt628:
- golang:alpine does not include 'git', see https://hub.docker.com/_/golang/
- in deference to https://github.com/coreos/container-linux-config-transpiler/pull/28#issuecomment-284691551
  Dockerfile builds a useful image that runs ct
- build-scratch with Dockerfile.build-scratch pares that down to just ct


dgonyeo:
Gonna let this sit for a day or so in case anyone else wants to comment on it

jkt628:
:+1:

dgonyeo:
@jkt628 this needs rebasing, I'll merge as soon as you can get around to that

jkt628:
done and tested.

is it possible to push the final image (coreos/ct) to hub.docker.com?  or, with this change, the [remote build triggers](https://docs.docker.com/docker-hub/builds/#remote-build-triggers) would automagically push the intermediate image.

dgonyeo:
I don't think I want to automagically push the image, unless it supports only pushing tagged releases. I'm not opposed to putting this on the docker hub though, I'll ask and see if that can happen.

jkt628:
you can choose the tag pushed into docker hub.  'latest' is expected to be just that, a build from the latest commit.  release tags are more difficult i think but do not consider me an expert.

perhaps 'latest' could always be the intermediate image and '_release_' could always be the final scratch image.

dgonyeo:
I would want `latest` to point to the latest release, not the latest commit. I could always just add "build a new docker image, put in docker hub" to the release process.

PR #36, opened 2017-03-09
config: refactored convert logic into config/types

dgonyeo:
As more sugar is added, the config/convert.go file was poised to
continually grow. This commit splits all conversion logic for each
config type into a function in the given type's file. These functions
are added to a list of converters at the beginning of runtime, which is
then iterated over to perform all conversions when the convert function
is called.

Upcoming PRs (can be collapsed into one if desired):
- refactor etcd logic to pass arguments via command line
- support for coreos-metadata


config/types/converter.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"reflect"
+
+	ignTypes "github.com/coreos/ignition/config/v2_0/types"
+	"github.com/coreos/ignition/config/validate"
+	"github.com/coreos/ignition/config/validate/report"
+)
+
+type converterFor2_0_0 func(in Config, out ignTypes.Config) (ignTypes.Config, report.Report)
```
crawford: Ignition doesn't use the patch number. Should be fine with 2_0 instead of 2_0_0.

dgonyeo: I was following the convention already in ct (ex [here](https://github.com/coreos/container-linux-config-transpiler/blob/master/config/convert.go#L34)), want me to drop the patch number in the couple of places outside this package too?

config/convert.go
```diff
@@ -15,345 +15,11 @@
...
-	"github.com/coreos/ignition/config/validate"
 	"github.com/coreos/ignition/config/validate/report"
-	"github.com/vincent-petithory/dataurl"
-)
-
-const (
-	BYTES_PER_SECTOR = 512
 )
 
 func ConvertAs2_0_0(in types.Config) (ignTypes.Config, report.Report) {
```
ajeddeloh: Should this exist if all it does is call a function with identical signature? Could we eliminate this and just have the caller call types.ConvertAs2_0_0 instead?

config/types/common.go
```diff
@@ -0,0 +1,60 @@
+// Copyright 2016 CoreOS, Inc.
```
ajeddeloh: 2017

config/types/converter.go
```diff
@@ -0,0 +1,58 @@
+// Copyright 2016 CoreOS, Inc.
```
ajeddeloh: 2017

config/types/converter.go
crawford: Yeah

config/convert.go
dgonyeo: I do like it being reexported in the `config` package for the sake of consumers of this package, since you'll also need the `Parse` function from it. Golang doesn't make reexporting functions easy though, which is why this looks like a very silly file.

crawford:
This really muddies the distinction between the public config package and the private internal package. If the types aren't something we want to export, we should pull them into internal. Otherwise, we should move the convert functions into internal.

dgonyeo:
I'll move the `types` package under `internal` and have the `config` package continue to reexport `ConvertAs2_0`, if that sounds reasonable?

crawford:
That convert function takes a type from the types package though. You could just mirror the structure of the files in internal and move the conversions there.

dgonyeo:
One thing I was going for with this refactor was keeping all logic related to a given type in the same place, but if you want me to move the convert functions into an internal package I can.

I'm also not sure this muddies much from a consumer of the API's perspective, there's now one more place you can call the convert function that you can ignore.

crawford:
Configuration and logic should never be together. That makes it very difficult for people to consume the config package without pulling in a bunch of dependencies.

PR #37, opened 2017-03-10
docs: rewritten, examples added, configuration extended, tests added

dgonyeo:
The existing docs have been rewritten
- README: now longer
- overview: rephrased, different sections
- configuration: format was altered
- configuration: now has an example config exercising every field

New docs have been added:
- examples: small, self-contained examples of common uses
- getting-started: walks the reader through creating a CLC and using ct

Tests have been added
- doc_test.go: finds all yaml blocks in doc/*.md, runs them through the
    parser and validator. Fails on any warnings or errors.

PR #39, opened 2017-03-13
config: templating implemented for etcd and flannel sections

dgonyeo:
(This PR is based on https://github.com/coreos/container-linux-config-transpiler/pull/38, so ignore all but the final commit. Will rebase once that PR is merged)

Users can now use `{}` to refer to dynamic information (like public ipv4
address) that will be filled in via an environment variable populated by
the coreos metadata service (which is also enabled and added as a
requirement if templating is used).

This is notably only usable inside the etcd and flannel sections. I could enable it in other places like systemd units, if we think it wouldn't be too confusing.

This commit also needs documentation to be written for it, but I'm holding off on doing that until https://github.com/coreos/container-linux-config-transpiler/pull/37 gets merged.

Example usage: https://gist.github.com/dgonyeo/cf500f91574fbc4441f6c346e43b89d0

Fixes https://github.com/coreos/bugs/issues/1821

### DIFF
--- a/config/templating/templating.go
+++ b/config/templating/templating.go
@@ -1,0 +1,128 @@
+// Copyright 2017 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package templating
+
+import (
+	"fmt"
+	"strings"
+)
+
+var (
+	ErrUnknownPlatform = fmt.Errorf("unsupported platform")
+	ErrUnknownField    = fmt.Errorf("unknown field")
+)
+
+const (
+	PlatformAzure  = "azure"
+	PlatformDO     = "digitalocean"
+	PlatformEC2    = "ec2"
+	PlatformGCE    = "gce"
+	PlatformPacket = "packet"
+)
+
+var Platforms = []string{
+	PlatformAzure,
+	PlatformDO,
+	PlatformEC2,
+	PlatformGCE,
+	PlatformPacket,
+}
+
+const (
+	fieldHostname  = "HOSTNAME"
+	fieldV4Private = "PRIVATE_IPV4"
+	fieldV4Public  = "PUBLIC_IPV4"
+	fieldV6Private = "PRIVATE_IPV6"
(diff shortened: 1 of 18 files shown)


config/templating/templating.go
```diff
@@ -0,0 +1,115 @@
...
+)
+
+var (
+	ErrUnknownPlatform = fmt.Errorf("unsupported platform")
+	ErrUnknownField    = fmt.Errorf("unknown field")
+)
+
+const (
+	PlatformAzure  = "azure"
+	PlatformDO     = "digital-ocean"
```
crawford: digitalocean

config/templating/templating.go
```diff
@@ -0,0 +1,115 @@
...
+	PlatformAzure,
+	PlatformDO,
+	PlatformEC2,
+	PlatformGCE,
+	PlatformPacket,
+}
+
+const (
+	fieldHostname  = "hostname"
+	fieldV4Private = "private_IPv4"
```
crawford: We should either make this all caps or all lowercase.

config/templating/templating.go
```diff
@@ -0,0 +1,115 @@
+// Copyright 2016 CoreOS, Inc.
```
ajeddeloh: 2017

config/templating/templating_test.go
```diff
@@ -0,0 +1,110 @@
+// Copyright 2016 CoreOS, Inc.
```
ajeddeloh: 2017

config/templating/templating_test.go
```diff
@@ -0,0 +1,110 @@
...
+			t.Errorf("#%d: hasTemplating didn't match", i)
+		}
+	}
+}
+
+func TestPerformTemplating(t *testing.T) {
+	tests := []struct {
+		platform     string
+		vars         []string
+		expectedVars []string
```
crawford: Please use the testing structure from Ignition: https://github.com/coreos/ignition/blob/8e66c946c57cca732041b1489492a1295c1ed0b1/config/translate_test.go#L34

dgonyeo:
nits addressed

dgonyeo:
Ah right, @dghubble: can you take a look at this and see if it has any important implications wrt matchbox? I'm fine tweaking this as needed, but it would be much easier to know now before this gets merged.

dghubble:
matchbox does templating before ct/fuze transpiles the config so if someone tried to add these metadata references, they'd need to be escaped there.

At the moment, fuze is still the vendored dependency and I don't expect this to break any reference clusters - the metadata service isn't used extensively and special etcd/flannel sections aren't used. I can't see much of a user case on bare-metal, but this shouldn't break things.

dgonyeo:
I just updated this to use `${var}` syntax instead of `$var` syntax. This required changing the templating logic.

May be worth another pass, but I just successfully brought up a box on ec2 with a single-node etcd cluster on it using this, so I'm pretty confident the changes were accurate. The changed logic also is covered by tests added here.

Currently blocking on the docs update PR so I can add documentation for this.

PR #40, opened 2017-03-13
 config/types/update: can now configure update group and server

dgonyeo:
(This PR is based on #36, so ignore all but the final commit. Will rebase once that PR is merged)

This commit lets users configure the update group and server for a machine:
```
update:
    channel: beta
    server: https://public.update.core-os.net/v1/update/
```

A resulting ignition config is viewable here (ignore the strategy bit): https://gist.github.com/dgonyeo/9f98a88353dbae078c66eb9487206919

There was no documentation on this config file that I could find, so someone should sanity check it for me.

Fixes https://github.com/coreos/bugs/issues/1861.

This new feature needs documentation, waiting on https://github.com/coreos/container-linux-config-transpiler/pull/37 to be merged before I write it.

### DIFF
--- a/config/types/update.go
+++ b/config/types/update.go
@@ -1,0 +1,88 @@
+// Copyright 2017 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package types
+
+import (
+	"errors"
+	"fmt"
+	"net/url"
+	"strings"
+
+	ignTypes "github.com/coreos/ignition/config/v2_0/types"
+	"github.com/coreos/ignition/config/validate/report"
+	"github.com/vincent-petithory/dataurl"
+)
+
+var (
+	ErrUnknownGroup = errors.New("unknown update group")
+)
+
+type Update struct {
+	Group  UpdateGroup  `yaml:"group"`
+	Server UpdateServer `yaml:"server"`
+}
+
+type UpdateGroup string
+type UpdateServer string
+
+func (u Update) Validate() report.Report {
+	switch strings.ToLower(string(u.Group)) {
+	case "stable", "beta", "alpha":
+		return report.Report{}
+	default:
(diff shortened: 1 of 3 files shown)


config/types/update.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+var (
+	ErrUnknownChannel  = errors.New("unknown update channel")
+	ErrUnknownStrategy = errors.New("unknown reboot strategy")
+)
+
+type Update struct {
+	Channel  UpdateChannel  `yaml:"channel"`
+	Strategy UpdateStrategy `yaml:"strategy"`
```
crawford: Let's go ahead and remove the strategy. This should be under a locksmith section instead.

dgonyeo: Removed. Want me to add a `locksmith` section in this PR, or another one?

crawford: Let's add it an another one.

config/types/update.go
```diff
@@ -0,0 +1,85 @@
+// Copyright 2016 CoreOS, Inc.
```
crawford: 2017

config/types/update.go
```diff
@@ -0,0 +1,85 @@
...
+	Channel UpdateChannel `yaml:"channel"`
+	Server  UpdateServer  `yaml:"server"`
+}
+
+type UpdateChannel string
+type UpdateServer string
+
+func (c UpdateChannel) Validate() report.Report {
+	switch strings.ToLower(string(c)) {
+	case "stable", "beta", "alpha":
```
crawford: They might be using a custom instance of CoreUpdate with different group names. We could warn if they haven't overridden the server and don't use one of the standard groups.

config/types/update.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/coreos/ignition/config/validate/report"
+	"github.com/vincent-petithory/dataurl"
+)
+
+var (
+	ErrUnknownChannel = errors.New("unknown update channel")
+)
+
+type Update struct {
+	Channel UpdateChannel `yaml:"channel"`
```
crawford: This should be a group, not a channel.

config/types/update.go
```diff
@@ -0,0 +1,85 @@
...
+				},
+			})
+		}
+		return out, report.Report{}
+	})
+}
+
+func updateConfContents(u *Update) string {
+	if u.Channel == "" {
+		u.Channel = "stable"
```
crawford: It should be safe to omit this and let the OS defaults kick in.

config/types/update_test.go
```diff
@@ -0,0 +1,44 @@
+// Copyright 2016 CoreOS, Inc.
```
crawford: 2017

dgonyeo:
nits addressed, PR updated to reflect the removal of reboot strategy

PR #42, opened 2017-03-20
config: added sugar for iptables

dgonyeo:
Allows configs to specify iptables rules such as:

```yaml
iptables:
    v4:
        output:
            default: "ACCEPT"
            rules:
                - -A INPUT -i lo -j ACCEPT
                - -A INPUT -i eth1 -j ACCEPT
                - -A INPUT -p tcp -m tcp --dport 222 -j ACCEPT
                - -A INPUT -m state --state RELATED,ESTABLISHED -j ACCEPT
                - -A INPUT -p ipv6-icmp -j ACCEPT
    v6:
        forward:
            default: "ACCEPT"
```

From this config the iptables save files are generated, and then
`ip(6)tables-store.service` and `ip(6)tables-restore.service` are enabled.

Fixes https://github.com/coreos/bugs/issues/1845 (and for that cc @euank)

One caveat is that once this section is enabled all chains default to `DENY`, and I don't know if it would make more sense to have the output chain default to `ACCEPT`.

Waiting on https://github.com/coreos/container-linux-config-transpiler/pull/37 being merged to write docs.

### DIFF
--- a/config/types/iptables.go
+++ b/config/types/iptables.go
@@ -1,0 +1,139 @@
+// Copyright 2017 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package types
+
+import (
+	"errors"
+	"fmt"
+
+	ignTypes "github.com/coreos/ignition/config/v2_0/types"
+	"github.com/coreos/ignition/config/validate/report"
+	"github.com/vincent-petithory/dataurl"
+)
+
+var (
+	ErrInvalidIptablesDefault = errors.New("iptables chain default must be one of ACCEPT or DROP")
+	ErrEmptyTableName         = errors.New("table cannot be empty")
+)
+
+type Iptables struct {
+	V4 []iptablesTable `yaml:"v4"`
+	V6 []iptablesTable `yaml:"v6"`
+}
+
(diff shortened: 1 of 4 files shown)


config/types/iptables.go
```diff
@@ -0,0 +1,121 @@
...
+	Forward iptablesRules `yaml:"forward"`
+	Output  iptablesRules `yaml:"output"`
+}
+
+type iptablesRules struct {
+	Default string   `yaml:"default"`
+	Rules   []string `yaml:"rules"`
+}
+
+func (ir iptablesRules) Validate() report.Report {
```
ajeddeloh: If we give Default it's own type (aliased to string) and define Validate() on that instead, it can give errors with greater precision.

config/types/iptables.go
```diff
@@ -0,0 +1,121 @@
...
+	case "ACCEPT", "DROP", "":
+		return report.Report{}
+	default:
+		return report.ReportFromError(ErrInvalidIptablesDefault, report.EntryError)
+	}
+}
+
+func init() {
+	register2_0(func(in Config, out ignTypes.Config) (ignTypes.Config, report.Report) {
+		if in.Iptables != nil {
```
ajeddeloh: I'd prefer this inverted, to return if in.IpTables == nil and prevent nesting

config/types/iptables.go
```diff
@@ -0,0 +1,121 @@
...
+					Enable: true,
+				})
+			}
+		}
+		return out, report.Report{}
+	})
+}
+
+func iptablesContents(chains *iptablesChains) ignTypes.FileContents {
+	contents := "*filter\n"
```
ajeddeloh: We should either specify in the docs that this section only configures the filter table or make this configurable. I'm in favor of the latter, since it might prevent needing to go in and change the spec later to accommodate it.

dgonyeo:
Added the ability to set the table, and added some documentation. Ready for another review pass.

dgonyeo:
Tweaked the PR, omittedd chains in the input config are now omitted from the produced iptables rules

dgonyeo:
I'm leaving Red Hat, and thus won't be getting this merged in. If someone else wants to open a new PR with these changes, feel free.

PR #44, opened 2017-03-21
config/types/locksmith: added configuration for locksmith reboot strat

dgonyeo:
Mostly a continuation of https://github.com/coreos/container-linux-config-transpiler/pull/40

Waiting on https://github.com/coreos/container-linux-config-transpiler/pull/37 to be merged before I write the docs

### DIFF
--- a/config/types/config.go
+++ b/config/types/config.go
@@ -22,15 +22,16 @@ import (
 )
 
 type Config struct {
-	Ignition Ignition `yaml:"ignition"`
-	Storage  Storage  `yaml:"storage"`
-	Systemd  Systemd  `yaml:"systemd"`
-	Networkd Networkd `yaml:"networkd"`
-	Passwd   Passwd   `yaml:"passwd"`
-	Etcd     *Etcd    `yaml:"etcd"`
-	Flannel  *Flannel `yaml:"flannel"`
-	Update   *Update  `yaml:"update"`
-	Docker   *Docker  `yaml:"docker"`
+	Ignition  Ignition   `yaml:"ignition"`
+	Storage   Storage    `yaml:"storage"`
+	Systemd   Systemd    `yaml:"systemd"`
+	Networkd  Networkd   `yaml:"networkd"`
+	Passwd    Passwd     `yaml:"passwd"`
+	Etcd      *Etcd      `yaml:"etcd"`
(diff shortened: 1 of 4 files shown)


config/types/locksmith.go
```diff
@@ -0,0 +1,41 @@
...
+
+type Locksmith struct {
+	RebootStrategy RebootStrategy `yaml:"reboot-strategy"`
+}
+
+type RebootStrategy string
+
+func (r RebootStrategy) Validate() report.Report {
+	switch strings.ToLower(string(r)) {
+	case "reboot", "etcd-lock", "best-effort", "off":
```
crawford: We've deprecated "best-effort". Go ahead and leave it out.

config/types/locksmith.go
```diff
@@ -0,0 +1,41 @@
...
+		return report.Report{}
+	default:
+		return report.ReportFromError(ErrUnknownStrategy, report.EntryWarning)
```
crawford: We can make this an error.

config/types/config.go
```diff
@@ -22,15 +22,16 @@ import (
...
+	Ignition  Ignition   `yaml:"ignition"`
+	Storage   Storage    `yaml:"storage"`
+	Systemd   Systemd    `yaml:"systemd"`
+	Networkd  Networkd   `yaml:"networkd"`
+	Passwd    Passwd     `yaml:"passwd"`
+	Etcd      *Etcd      `yaml:"etcd"`
+	Flannel   *Flannel   `yaml:"flannel"`
+	Update    *Update    `yaml:"update"`
+	Docker    *Docker    `yaml:"docker"`
+	Locksmith *Locksmith `yaml:"locksmith"`
```
crawford: Why make this a pointer? I know the others do it as well (and the question stands for them as well).

dgonyeo: They're pointers so we can easily tell when the user omits the section. It's not particularly important I suppose, but it's so I can catch this case:

```yaml
locksmith:
    strategy:
```

dgonyeo: I originally made the tweak in Andrew's PR that added etcd and flannel, the validation logic was choking when the section was omitted because `""` isn't a valid version for etcd. Making the value `nil`-able by making it a pointer made the validation logic for the section not run when the section was omitted.

PR #53, opened 2017-03-31
main: only print reports if there are entries

crawford:
Without this change, two newlines are emitted on stderr, which offsets
the output of ct when run in a terminal.


dgonyeo:
needs rebase

PR #58, opened 2017-04-17
config/types: override coreos-metadata provider on openstack-metadata

dgonyeo:
The provider as specified by the kernel command line to coreos-metadata
must be overridden when using the openstack-metadata provider, as the
container linux image merely specifies "openstack" due to it not knowing
which metadata strategy will be used.

Fixes https://github.com/coreos/bugs/issues/1917

### DIFF
--- a/config/types/common.go
+++ b/config/types/common.go
@@ -18,12 +18,30 @@ import (
 	"fmt"
 	"reflect"
 
+	ignTypes "github.com/coreos/ignition/config/v2_0/types"
+	"github.com/coreos/ignition/config/validate/report"
+
 	"github.com/coreos/container-linux-config-transpiler/config/templating"
 )
 
 var (
 	ErrPlatformUnspecified = fmt.Errorf("platform must be specified to use templating")
 )
+
+func init() {
+	register2_0(func(in Config, out ignTypes.Config, platform string) (ignTypes.Config, report.Report) {
+		if platform == templating.PlatformOpenStackMetadata {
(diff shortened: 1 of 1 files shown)


dgonyeo:
One question I have is should ct always override the coreos-metadata provider on this platform, or only when dynamic data is used? This PR currently does the latter.

crawford:
> One question I have is should ct always override the coreos-metadata provider on this platform, or only when dynamic data is used?

The platform is only needed if templating is used. I would apply the override whenever the platform is specified; regardless of whether or not templating is used.

config/types/etcd.go
```diff
@@ -126,13 +128,16 @@ func init() {
 					Contents: contents,
 				}},
 			})
+			if hasTemplating && platform == templating.PlatformOpenStackMetadata {
```
crawford: I don't like that we have to remember to include this everywhere templating is used. Is there anyway to abstract this so that it's handled in one place. That would also relieve the need for that hideous global state variable.

dgonyeo:
I just put a little more thought into this and came up with a much more elegant solution.

Also tweaked the behavior so now it always runs when openstack-metadata is specified, regardless of the use of dynamic data.

dgonyeo:
I've made a PR to coreos-metadata to allow provider overriding via an environment variable:
https://github.com/coreos/coreos-metadata/pull/38

I've also updated this PR to use those changes.

dgonyeo:
This has been updated to reflect the changes that slipped into https://github.com/coreos/coreos-overlay/pull/2524 before it was merged.

PR #59, opened 2017-04-17
internal: remove "generated ignition config is invalid" line

dgonyeo:
If conversion fails, the report is printed and the program is exited.
Before this commit the string "Generated Ignition config was invalid."
was printed to stderr, which is unnecessary since the report was already
printed.


dgonyeo:
As an additional example of why we don't want this line printed, checking if a platform was specified for dynamic data doesn't happen unitl conversion, because the presence of dynamic data isn't checked until then. This error is entirely unrelated to the generated ignition config, and is described sufficiently by the report that gets printed.

PR #60, opened 2017-04-17
config/config.go: add comments for the exported functions

dgonyeo:
Doing this because @dghubble was confused about an added argument to `ConvertAs2_0`, and if these comments are in place I'll remember to update them if fields are added in the future.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -24,6 +24,9 @@ import (
 	"github.com/coreos/ignition/config/validate/report"
 )
 
+// Parse will convert a byte slice containing a Container Linux Config into a
+// golang struct representing the config, and a report of any warnings or errors
+// that occurred during the parsing.
 func Parse(data []byte) (types.Config, report.Report) {
 	var cfg types.Config
 	var r report.Report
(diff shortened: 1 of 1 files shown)


dghubble:
Enabling golint will prevent all manner of these missing comments.

config/config.go
```diff
@@ -24,6 +24,9 @@ import (
 	"github.com/coreos/ignition/config/validate/report"
 )
 
+// Parse will convert a byte slice containing YAML (a container linux config)
```
crawford: Drop the mention of YAML. It's just a Container Linux Config.

config/config.go
```diff
@@ -54,6 +57,11 @@ func Parse(data []byte) (types.Config, report.Report) {
 	return cfg, r
 }
 
+// ConvertAs2_0 will convert a golang struct representing a container linux
```
crawford: Capitalize Container Linux Config and Ignition Config. These are proper nouns.

dgonyeo:
nits addressed

PR #62, opened 2017-04-23
Add Makefile to build static bin/ct binary

dghubble:
`ct` release binaries are dynamically linked and don't work in various containers (e.g. scratch, alpine). Let's define a Makefile to formalize the build process so its not as dependent on whatever env vars the release engineer happens to have set. Notably, set CGO_ENABLED to 0.

I've left the old atypical build script which does symlinking, pending comment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,21 @@
+export CGO_ENABLED:=0
+
+VERSION=$(shell git describe --dirty)
+LD_FLAGS="-w -X github.com/coreos/container-linux-config-transpiler/version.Raw=$(VERSION)"
+
+REPO=github.com/coreos/container-linux-config-transpiler
+
+all: build
+
+build: bin/ct
+
+bin/ct:
+	@go build -o $@ -v -ldflags $(LD_FLAGS) $(REPO)/internal
+
+test:
+	@./test
+
+clean:
+	@rm -rf bin
+
+.PHONY: all build clean test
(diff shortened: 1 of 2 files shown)


test
```diff
@@ -29,6 +29,6 @@ echo "Checking govet..."
 go vet $PKG_VET
 
 echo "Running tests..."
-go test -timeout 60s -cover $@ ${PKG} --race
+go test -timeout 60s -cover $@ ${PKG}
```
dgonyeo: Why remove the race flag?

dghubble: go test: -race requires cgo; enable cgo by setting CGO_ENABLED=1

dgonyeo: :+1:

dgonyeo:
Oh wait, before I merge this:

Do we want to keep the build bash script in the repo? Perhaps that should be removed if we're adding this Makefile?

Makefile
```diff
@@ -0,0 +1,21 @@
...
+LD_FLAGS="-w -X github.com/coreos/container-linux-config-transpiler/version.Raw=$(VERSION)"
+
+REPO=github.com/coreos/container-linux-config-transpiler
+
+all: build
+
+build: bin/ct
+
+bin/ct:
+	@go build -o bin/ct -v -ldflags $(LD_FLAGS) $(REPO)/internal
```
crawford: `-o $@`

dghubble: Added

dghubble:
@dgonyeo I support its removal, but there was resistance to setting up a GOPATH in the past that meant the symlink trick was kept. Proceed with removing the script based build?

PR #63, opened 2017-04-23
Add glide.yaml manifest and Makefile vendor target

dghubble:
Declare dependencies using a glide.yaml manifest. Simply declare dependencies in `glide.yaml` like you were doing manually in `vendor.manifest` and run `make vendor`.

* Glide provides a repeatable vendor process and its the defactor standard (see etcd, dex, matchbox, tectonic, k8s projects, etc.)
* Friendly for upstream projects which depend on ct. Before, each ct dependency had to be vendored one by one, by hand. Inventing a vendor manifest means tools can't parse it.


Makefile
```diff
@@ -0,0 +1,21 @@
...
+LD_FLAGS="-w -X github.com/coreos/container-linux-config-transpiler/version.Raw=$(VERSION)"
+
+REPO=github.com/coreos/container-linux-config-transpiler
+
+all: build
+
+build: bin/ct
+
+bin/ct:
+	@go build -o bin/ct -v -ldflags $(LD_FLAGS) $(REPO)/internal
```
crawford: `- $@`

dghubble:
glide-vc (from @sgotti) is used to remove vendored docs, unused subpackages, and other code which isn't needed. If you're unhappy with that, the vc step can be removed (though bootkube, dex, etcd, tectonic, and matchbox, etc. use it). 19,000 lines are removed because 19000 lines were vendored by hand and aren't required.

dghubble:
I think the reason we've gravitated toward the Makefile in those projects is that it alleviates the need to figure out each project's custom scripts setup and oddities when working with each one. Even if some tasks just call your existing scripts (such as test). I think the concepts of named targets, dependent tasks, and the desire to use a common tool has mapped well, especially with more contributors.

crawford:
Huh, well if the tests are passing, I guess it didn't remove too much. I'm still not a fan of the Makefile, but I don't feel too strongly about it. I defer to @dgonyeo.

dghubble:
Yeah, this has been a bit of a concerted effort by various folks to get `make build`, `make test`, `make install` to be easy, predicatable commands across projects. It'd be nice to have CL projects get the same benefits if you're a fan @dgonyeo?

* https://github.com/coreos/dex/blob/master/Makefile
* https://github.com/coreos/matchbox/blob/master/Makefile
* https://github.com/coreos/tectonic-installer/blob/master/installer/Makefile
* https://github.com/kubernetes-incubator/bootkube/blob/master/Makefile

crawford:
For what it's worth, Ignition used to use a Makefile, but @vcaputo and I agreed that it was silly to shoehorn a script into a Makefile. There is something to be said for consistency though.

dghubble:
For sure, definitely pros and cons to each. Within Tectonic, by 3000 commits we had a lot of scripts lying around and make enforced some discipline and did a lot of what scripts tried to do for free.

Makefile
dghubble: added

dgonyeo:
I don't have a particularly strong opinion here, let's go for it for consistency's sake

dgonyeo:
@dghubble the hash commits for the commit shared between this and https://github.com/coreos/container-linux-config-transpiler/pull/62 don't match, which commit should be merged in?

dghubble:
#62 should be first since it adds the Makefile. I've rebased this now too, so the commit here matches #62

PR #64, opened 2017-04-25
travis.yml: Add test coverage for Go 1.7 and 1.8

dghubble:
I think we mean to be supporting and testing against recent Go versions. Leave Go 1.5 or drop?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,8 @@
 language: go
-matrix:
-  include:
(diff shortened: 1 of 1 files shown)


dgonyeo:
If you can also drop go 1.5 and 1.6 in this that'd be great

dghubble:
Sure, dropped.

dghubble:
Rebased

PR #66, opened 2017-05-02
config/types: use %q for 'string' values

gyuho:
boolean, integer values are broken in %q

We are getting

```
key, val: auto-compaction-retention '\x01'
key, val: debug %!q(bool=true)
```

since the default format string `"--%s=%q"`.

If there's any better way to handle this, please feel free to close this :)

Thanks!

### DIFF
--- a/config/types/common.go
+++ b/config/types/common.go
@@ -97,6 +97,10 @@ func getArgs(format, tagName string, e interface{}) []string {
 				vars = append(vars, getCliArgs(val)...)
 			} else {
 				key := et.Field(i).Tag.Get(tagName)
+				if _, ok := val.(string); ok {
+					// to handle whitespace characters
+					val = fmt.Sprintf("%q", val)
+				}
 				vars = append(vars, fmt.Sprintf(format, key, val))
 			}
 		}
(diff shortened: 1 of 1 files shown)


dgonyeo:
`%q` was used so that spaces in arguments would be handled, maybe special casing for strings is needed here?

george-angel:
Getting the same issue:

My etcd-member is failing to start with `etcd-wrapper[1686]: invalid boolean value "%!q(bool=true)" for -peer-client-cert-auth: strconv.ParseBool: pars...`, and sure enough I have a bool for that vol in my config, changing it to a string errors with: `line 265: cannot unmarshal !!strtrueinto bool`

Building with this change fixes the issue.

dgonyeo:
@gyuho yeah, exactly

gyuho:
@dgonyeo Fixed. PTAL.

e.g.

```
key, val: name "s1"
key, val: data-dir "/var/lib/etcd"
key, val: listen-peer-urls "http://localhost:2380"
key, val: listen-client-urls "http://localhost:2379"
key, val: initial-advertise-peer-urls "http://localhost:2380"
key, val: initial-cluster "s1=http://localhost:2380"
key, val: initial-cluster-state "new"
key, val: initial-cluster-token "mytoken"
key, val: advertise-client-urls "http://localhost:2379"
key, val: auto-compaction-retention 1
key, val: debug true
```

dgonyeo:
awesome, thanks!

george-angel:
Is there a way to track when there is a release with this fix? Thank you

PR #70, opened 2017-05-09
config/types/etcd: add warnings when some fields don't start with schema

dgonyeo:
Some fields in etcd must include schemas in the given URL (http, https,
unix, or unixs), this commit adds a validation field on the options
structs that will exmit warnings if these fields don't appear to start
with the correct values.

Related to https://github.com/coreos/bugs/issues/1952#issuecomment-299489651

### DIFF
--- a/config/types/etcd.go
+++ b/config/types/etcd.go
@@ -17,6 +17,7 @@ package types
 import (
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/coreos/go-semver/semver"
 	ignTypes "github.com/coreos/ignition/config/v2_0/types"
@@ -32,7 +33,9 @@ var (
 )
 
 // Options can be the options for any Etcd version
-type Options interface{}
+type Options interface {
+	Validate() report.Report
+}
 
 type etcdCommon Etcd
 
@@ -147,6 +150,47 @@ func etcdContents(etcd Etcd, platform string) (string, error) {
 	return unit.String(), nil
 }
 
+var validPrefixes = []string{"http://", "https://", "unix://", "unixs://"}
+
+func checkForPrefix(s string) bool {
+	if s == "" {
+		return true
+	}
+	urlStrs := strings.Split(s, ",")
+	for _, urlStr := range urlStrs {
+		hasPrefix := false
+		for _, p := range validPrefixes {
+			hasPrefix = hasPrefix || strings.HasPrefix(urlStr, p)
+		}
+		if !hasPrefix {
+			return false
+		}
+	}
+	return true
+}
+
+func validatePrefixes(m map[string]string) report.Report {
+	r := report.Report{}
+	for name, val := range m {
+		if !checkForPrefix(val) {
+			r.Add(report.Entry{
+				Message: fmt.Sprintf("value for %s doesn't start with one of: %v", name, validPrefixes),
+				Kind:    report.EntryWarning,
+			})
+		}
+	}
+	return r
+}
+
+func (e Etcd3_0) Validate() report.Report {
+	return validatePrefixes(map[string]string{
+		"listen_peer_urls":            e.ListenPeerUrls,
+		"listen_client_urls":          e.ListenClientUrls,
+		"initial_advertise_peer_urls": e.InitialAdvertisePeerUrls,
+		"advertise_client_urls":       e.AdvertiseClientUrls,
+	})
+}
+
 type Etcd3_0 struct {
 	Name                     string `yaml:"name"                        cli:"name"`
 	DataDir                  string `yaml:"data_dir"                    cli:"data-dir"`
@@ -192,6 +236,15 @@ type Etcd3_0 struct {
 	Debug                    bool   `yaml:"debug"                       cli:"debug"`
 	LogPackageLevels         string `yaml:"log_package_levels"          cli:"log-package-levels"`
 	ForceNewCluster          bool   `yaml:"force_new_cluster"           cli:"force-new-cluster"`
+}
+
+func (e Etcd3_1) Validate() report.Report {
+	return validatePrefixes(map[string]string{
+		"listen_peer_urls":            e.ListenPeerUrls,
+		"listen_client_urls":          e.ListenClientUrls,
+		"initial_advertise_peer_urls": e.InitialAdvertisePeerUrls,
+		"advertise_client_urls":       e.AdvertiseClientUrls,
+	})
 }
 
 type Etcd3_1 struct {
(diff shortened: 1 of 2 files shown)


config/types/etcd.go
```diff
@@ -142,6 +145,39 @@ func etcdContents(etcd Etcd, platform string) (string, error) {
 	return assembleUnit("/usr/lib/coreos/etcd-wrapper $ETCD_OPTS", args, vars, platform)
 }
 
+var validPrefixes = []string{"http://", "https://", "unix://", "unixs://"}
+
+func checkForPrefix(s string) bool {
+	hasPrefix := false
+	for _, p := range validPrefixes {
+		hasPrefix = hasPrefix || strings.HasPrefix(s, p)
```
crawford: Use the URL package instead.

config/types/etcd.go
```diff
@@ -32,7 +33,9 @@ var (
 )
 
 // Options can be the options for any Etcd version
-type Options interface{}
+type Options interface {
```
crawford: Why is this being added?

config/types/flannel.go
```diff
@@ -113,6 +113,10 @@ func flannelContents(flannel Flannel, platform string) (string, error) {
 	return assembleUnit("/usr/lib/coreos/flannel-wrapper $FLANNEL_OPTS", args, vars, platform)
 }
 
+func (f Flannel0_7) Validate() report.Report {
```
crawford: Why are these being added?

config/types/etcd.go
```diff
@@ -142,6 +145,39 @@ func etcdContents(etcd Etcd, platform string) (string, error) {
...
+
+func (e Etcd3_0) Validate() report.Report {
+	r := report.Report{}
+	addPrefixWarning := func(name string) {
+		r.Add(report.Entry{
+			Message: fmt.Sprintf("%s doesn't start with one of: %v", name, validPrefixes),
+			Kind:    report.EntryWarning,
+		})
+	}
+	if !checkForPrefix(e.ListenPeerUrls) {
```
crawford:
```go
for key, attr in map[string]string {
    "listen_peer_urls":            e.ListenPeerUrls,
    "listen_client_urls":          e.ListenClientUrls,
    "initial_advertise_peer_urls": e.InitialAdvertisePeerUrls,
    "advertise_client_urls":       e.AdvertiseClientUrls,
} {
    if !checkForPrefix(attr) {
        r.Add(report.Entry{
            Message: fmt.Sprintf("value for %s doesn't start with one of: %v", prefix, validPrefixes),
            Kind:    report.EntryWarning,
        })
    }
}
```

You probably want to put that in its own function to avoid duplication between the different Validate methods.

config/types/etcd.go
dgonyeo: This is necessary for the reflection logic to properly check this field for a `Validate` function. Without this the added validation logic to check for the schema prefixes is never called.

config/types/flannel.go
dgonyeo: These are necessary for them to still implement the `Options` interface

config/types/etcd.go
dgonyeo: The URL package chokes on dynamic data:
```
parse http://{PRIVATE_IPV4}:2380: invalid character "{" in host name
```

dgonyeo:
Refactored a little, got tests passing, and got it to handle comma separated lists of URLs

config/types/etcd.go
```diff
@@ -32,7 +33,9 @@ var (
 )
 
 // Options can be the options for any Etcd version
```
crawford: Not part of this PR, but "etcd" is always lowercase.

dgonyeo: I blame @ajeddeloh for that

config/types/etcd.go
crawford: That doesn't seem right. The validation code already has this interface declared and uses that to do type-casts.

config/types/etcd.go
```diff
@@ -147,6 +150,47 @@ func etcdContents(etcd Etcd, platform string) (string, error) {
...
+	}
+	return true
+}
+
+func validatePrefixes(m map[string]string) report.Report {
+	r := report.Report{}
+	for name, val := range m {
+		if !checkForPrefix(val) {
+			r.Add(report.Entry{
+				Message: fmt.Sprintf("%s doesn't start with one of: %v", name, validPrefixes),
```
crawford: The field name will never start with one of those prefixes. "value for %s doesn't..."

config/types/etcd.go
```diff
@@ -147,6 +150,47 @@ func etcdContents(etcd Etcd, platform string) (string, error) {
...
+
+func checkForPrefix(s string) bool {
+	if s == "" {
+		return true
+	}
+	urlStrs := strings.Split(s, ",")
+	for _, urlStr := range urlStrs {
+		hasPrefix := false
+		for _, p := range validPrefixes {
+			hasPrefix = hasPrefix || strings.HasPrefix(urlStr, p)
```
crawford: I asked before. Use the URL package.

dgonyeo: The URL package chokes on dynamic data:
```
parse http://{PRIVATE_IPV4}:2380: invalid character "{" in host name
```

config/types/etcd.go
dgonyeo: I didn't dig into the implementation of the validation logic, but the code wasn't called without this change. I can spend some time digging into the validation logic if you'd prefer me to try to fix it in there, but I have a feeling it's due to it not being able to know what functions are available to call on an empty interface.

crawford: But that's how Go works. All of this is determined at runtime instead of at compile time. You don't specify which interfaces you are implementing (e.g. Rust's `impl`), it uses duck-typing to figure out which interfaces are implicitly satisfied. Maybe that is broken across packages.

config/types/etcd.go
crawford: I'm not familiar with the full URL RFC. You'll need to review that before this is safe to do. Otherwise, substitute a dummy IP address before validation.

config/types/etcd.go
ajeddeloh: whoops.

config/types/etcd.go
ajeddeloh: This is hitting a weird edge case with the validation code. The validation code "flattens" all embedded structs, so its checking if you can call Validate() on the top level `Etcd` struct. Go's struct embeddeding rules allows that if an embedded member allows that, but only superficially (doesn't work with []T or map[T1]T2). Unfortunately (for us), go uses the interface supplied not the underlying data's types. 

Why not use the Validate<NAME>() style functions? The flattening *should* handle that just fine.

dgonyeo:
I'm leaving Red Hat, and thus won't be getting this merged in. If someone else wants to open a new PR with these changes, feel free.

PR #71, opened 2017-05-12
Makefile: Add release target to produce static binaries

dghubble:
I noticed the v0.3.1 release is still dynamically linked and was build with the build script. Could we formalize the Makefile build/release?

Run `make release` at a release tag (say v0.3.2) to produce:

```
ls _output
ct-v0.3.2-x86_64-apple-darwin 
ct-v0.3.2-x86_64-pc-windows-gnu
ct-v0.3.2-x86_64-unknown-linux-gnu
```

I've matched the existing names (unknown-linux-gnu, apple-darwin, pc-windows-gnu), though they seem odd. Is there a reason for those? Maybe just linux, darwin, windows? cc @crawford

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 bin/
 gopath/
 *.swp
+_output/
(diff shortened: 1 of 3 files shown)


dghubble:
Leaving the build script for now. Test script is written to use it so I think that's a separate change from adding a release target.

PR #72, opened 2017-05-15
docs: update code fences to use clcs

crawford:
This will allow these blocks to render properly on coreos.com.

### DIFF
--- a/doc/examples.md
+++ b/doc/examples.md
@@ -4,7 +4,7 @@ Here you can find a bunch of simple examples for using ct, with some explanation
 
 ## Users and groups
 
-```yaml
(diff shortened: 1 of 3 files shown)


doc/examples.md
```diff
@@ -4,7 +4,7 @@ Here you can find a bunch of simple examples for using ct, with some explanation
 
 ## Users and groups
 
-```yaml
+```container-linux-config
```
dghubble: Fancy

crawford: Actually, I'm going to change this to `yaml container-linux-config`. That way, both GitHub and our site render it properly.

PR #77, opened 2017-05-23
config/types: add etcd v3.2

gyuho:
We are planning to release v3.2 in one or two week(s).
Pushing it in advance for review.


crawford:
I think this is easier to follow as two commits as it was originally done.

gyuho:
Can we merge this? 3.2 is just released https://github.com/coreos/etcd/releases/tag/v3.2.0.

Thanks.

PR #78, opened 2017-06-05
main: expand strict coverage

crawford:
When the strict flag was originally introduced, it didn't take into
account the fact that the conversion function can return non-fatal
reports. This corrects that mistake by failing when there are any
reports that are returned.


dgonyeo:
Whoops, that report is for the parsing logic, and this report is for the validation logic. nvm.

PR #86, opened 2017-06-21
config/types: override sshkeys's opts on openstack-metadata

dgonyeo:
coreos-metadata-sshkey@.service's provider should be overridden on the
openstack-metadata platform. This commit makes that change.


crawford:
Can you shorten the commit title so that it fits?

PR #90, opened 2017-07-14
build: fix version handling

AlexNPavel:
The old Makefile had a typo that caused the output binaries to
have the version unchanged ("ct was not built properly").
This also removes the build script as it was unneccessary and
update the test file, which previously depended on that script.


test
```diff
@@ -1,35 +1,28 @@
-#!/usr/bin/env bash
```
crawford: Use `/usr/bin/env bash`. Not all distros keep bash in `/bin`.

test
```diff
@@ -1,35 +1,28 @@
...
+#!/bin/bash -e
```
dghubble: I'd keep the old version of this line. Also don't need to have -e and then set -e right below.

test
```diff
@@ -1,35 +1,28 @@
...
 
-PKG=$(cd gopath/src/${REPO_PATH}; go list ./... | \
-	grep --invert-match vendor)
-
-# https://github.com/golang/go/issues/15067
-PKG_VET=$(cd gopath/src/${REPO_PATH}; go list ./... | \
-	grep --invert-match vendor | \
-	grep --invert-match internal/log)
+PKGS=$(go list ./... | grep -v /vendor)
+FORMATTABLE=$(ls -d */ | grep -v -E '(vendor/|examples/)')
```
dghubble: You can remove the exclusion of examples since there are is no examples folder in this project.

test
```diff
@@ -1,35 +1,28 @@
...
 
 echo "Checking gofix..."
 go tool fix -diff $SRC
 
 echo "Checking gofmt..."
```
dghubble: were you going to add an issue or TODO to add `golint` checking? Its the only one missing

PR #92, opened 2017-07-14
build: various fixes to the build programs

crawford:


Makefile
```diff
@@ -1,41 +1,50 @@
...
+export GOPATH=$(shell pwd)/gopath
+export CGO_ENABLED:=0
 
-build: bin/ct
+.PHONY: all
+all: bin/ct
 
-bin/ct:
-	@go build -o $@ -v -ldflags $(LD_FLAGS) $(REPO)/internal
+gopath:
```
dghubble: I'd prefer to see all the funny business with gopath symlinking gone. Setting up a GOPATH is step one of installing Go.

dghubble:
```
crawford:
    $(Q)mkdir -p gopath/src/github.com/coreos
    $(Q)ln -s ../../../.. gopath/src/$(REPO)
```
:P

crawford: As mentioned in the commit message, I'm not a fan of the global GOPATH and I don't use it on my system. There is also testing benefit to using a separate GOPATH, since you won't pull in anything from the surrounding environment.

PR #98, opened 2017-08-07
glide.yaml: vendor ignition v0.17.2

dghubble:
We'd like to update ct within matchbox since it hasn't been updated in a while https://github.com/coreos/matchbox/issues/606. Using Ignition v0.17.2 seems like a good place to make the cut.

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: 108dfecb065ae7628c6eea011e78e9ae25fc029d8770d6784ad26ede77a90ba5
-updated: 2017-06-20T18:10:38.968919013-04:00
+hash: 866f7f8e6abd6c52e124db97d9c5665e261f65a06289e9803dfd18c6720ca40f
+updated: 2017-08-07T13:11:10.67669909-07:00
 imports:
 - name: github.com/ajeddeloh/yaml
   version: 1072abfea31191db507785e2e0c1b8d1440d35a5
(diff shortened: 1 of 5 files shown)


dghubble:
Yeah, that would be a problem if it did. Matchbox calls `ct.Parse(bytes)` and the docs don't quite specify.  Inspecting, it looks like `ConvertAs2_0` is [available](https://github.com/coreos/container-linux-config-transpiler/blob/master/config/config.go#L70) and tests do equality testing against v2.0 still. https://github.com/coreos/container-linux-config-transpiler/blob/master/config/config_test.go#L560 But I may be wrong @dgonyeo 

This may be a bit of a chicken and egg problem, because the ct package won't get usage by Tectonic, except through matchbox vendoring it

dgonyeo:
`ct` [imports](https://github.com/coreos/container-linux-config-transpiler/blob/d2f7243130995a6f0d545c9a7651df582a89cf1e/config/types/converter.go#L21) the `v2_0/types` package from Ignition, so the [returned ignTypes.Config](https://github.com/coreos/container-linux-config-transpiler/blob/d2f7243130995a6f0d545c9a7651df582a89cf1e/config/types/converter.go#L34) will be a 2.0.0 config, not a 2.1.0 config. I just checked out this PR and built it to double check, and `ct` still produces 2.0.0 configs with this change.

dghubble:
Rebased per "updated branch" requirement.

PR #99, opened 2017-08-07
travis: Build & push docker image to quay if on master

chancez:


dgonyeo:
I think I'd prefer to only push up releases instead of every successful build, so it'd maybe be a target of the `Makefile` or something. I could be talked into every release though.

chancez:
@dgonyeo Yeah, I was trying to think of how to do that, but since your release flow isn't in travis that's a lot bigger of a change. At the least this should get it published in a container image form, and I think if you wanted to optimize that, you could later.

chancez:
Btw I'm not blocked by this, it just seemed super easy to get a minimal thing building/pushing. I mostly want the image for easy testing locally since i prefer to use docker where I can.

crawford:
Does it make sense to push it to a different tag? We don't want people to pull a potentially broken image if they are blindly pulling latest. Maybe "latest-dev"?

chancez:
Travis doesn't seem to be picking up my .travis.yml changes, and seemingly think's it's current branch is master. I'm not sure what's up. Could certainly push to a different tag.

dghubble:
We have several projects which push an image for each commit to master. The images are tagged with the Git SHA as per the git-version script to allow users to use an image at a specific commit if truly desired.

chancez:
It's picking up changes now, but still think's it's on master (it's still trying to run docker login/push), and for some reason isn't able to access the docker username/password environment variables.

I combined your suggested changes. It now tags with the git sha and latest-dev as the docker image tags.

dgonyeo:
fwiw I'm happy with this now

build-scratch
```diff
@@ -1,4 +1,8 @@
 #!/bin/sh
-SHA=`git rev-parse HEAD`
-docker build -t $SHA . &&
-    docker run --rm --entrypoint tar $SHA -c /usr/bin/ct Dockerfile.build-scratch | docker build -f Dockerfile.build-scratch --tag coreos/ct -
+
+set -e
+VERSION=$(./git-version)
```
crawford: Please don't use the git-version script. I thought we got rid of that thing. `git describe` is sufficient.

build-scratch
```diff
@@ -1,4 +1,8 @@
...
+docker build -t $VERSION .
+docker run --rm --entrypoint tar $VERSION -c /usr/bin/ct Dockerfile.build-scratch \
+    | docker build -f Dockerfile.build-scratch --tag quay.io/coreos/ct:$VERSION -
+docker tag quay.io/coreos/ct:$VERSION quay.io/coreos/ct:latest-dev
```
crawford: Why are we tagging the container with the hash?

build-scratch
chancez: Can you elaborate? How does one do the equivlent of getting the git tag or commit sha with `git describe`?

build-scratch
chancez: It's the sha1/tag based on git-version, there isn't any hash unless I'm misunderstanding. This just tags it with the sha, and the latest-dev tag.

build-scratch
crawford:
```
$ git describe
v0.5.0-21-g66f2e35
```

If it's a tagged commit, `git describe` will just give you the tag.

build-scratch
crawford: When I say "hash", I'm referring to the sha1. I see now though, that by pushing the hash/tag, we will automatically get tagged images instead of having to do something different for the release process. This seems fine.

build-scratch
chancez: @crawford If I use git describe locally I get the following error:
```
fatal: No names found, cannot describe anything.
```

crawford: :man_shrugging: Works fine for me and the rest of our build system.

```
$ git describe
v0.4.2
```

chancez: Ah, I was running it in another repo. Seems this fails if you don't have any tags yet. I suppose you don't care about that case?

chancez:
Ok I've updated it. I can rebase after I get a final approval before it's merged if you prefer.

chancez:
Rebased

crawford:
It would be nice to also have this push tagged releases to the "latest" quay tag.

chancez:
Could be done in a follow up probably.

lubo:
What's the status here? Having the image would be really helpful.

sdemos:
@chancez I would like to merge this for the next release. Is it ready to go? Also, can you rebase it to pull in the latest Dockerfile changes?

stuart-warren:
@chancez if you don't want to rebase, I will. I'd really like this too.

chancez:
Go for it!

stuart-warren:
Have created a new merge request to supercede this #140

dghubble:
Yeah, let's go with #140

PR #100, opened 2017-08-08
news: add news for v0.4.2

dgonyeo:


NEWS
```diff
@@ -1,3 +1,11 @@
+08-Aug-2017 CONFIG-TRANSPILER v0.4.2
+
+  Fixes
+
+    - Error when an unsupported platform is used
```
crawford: Error is not a verb.

dgonyeo: Changed to "Emit an error..."

PR #102, opened 2017-08-14
platforms: change vagrant-virtualbox to virtualbox

AlexNPavel:
As part of using generic oem-ids for vagrant provisioned VMS, we are
changing vagrant-virtualbox to virtualbox.

### DIFF
--- a/config/platform/platform.go
+++ b/config/platform/platform.go
@@ -21,7 +21,7 @@ const (
 	GCE               = "gce"
 	Packet            = "packet"
 	OpenStackMetadata = "openstack-metadata"
-	VagrantVirtualbox = "vagrant-virtualbox"
+	Virtualbox        = "virtualbox"
 )
 
 var Platforms = []string{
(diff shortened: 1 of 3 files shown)


dgonyeo:
`vagrant-virtualbox` made it into releases, so I think we should alias `vagrant-virtualbox` to `virtualbox`. While it makes me feel a little icky, maybe the easiest way to do that would be:
```go
func init() {
    platformTemplatingMap[platform.VagrantVirtualbox] = platformTemplatingMap[platform.Virtualbox]
}
```

sdemos:
what is the status of this change? are we still planning on changing the name of this oem?

 /cc @crawford @dgonyeo

crawford:
I forget the status. @bgilbert would know.

bgilbert:
We will be changing the OS and Ignition OEM name to `virtualbox`, but not the CT platform name.

sdemos:
In that case I'll close this pr.

PR #103, opened 2017-08-23
docs: document -hasroot suffix for root on raid

ajeddeloh:
The initramfs will only start raid arrays with -hasroot. Without this the root partition cannot be found.

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -29,7 +29,7 @@ _Note: all fields are optional unless otherwise marked_
       * **start** (string): the start of the partition with a unit (KiB, MiB, GiB). If zero, the partition will be positioned at the earliest available part of the disk.
       * **type_guid** (string): the GPT [partition type GUID][part-types]. If omitted, the default will be 0FC63DAF-8483-4772-8E79-3D69D8477DE4 (Linux filesystem data).
   * **raid** (list of objects): the list of RAID arrays to be configured.
-    * **name** (string, required): the name to use for the resulting md device.
+    * **name** (string, required): the name to use for the resulting md device. If the array will contain the root partition, it must be suffixed with "-hasroot".
     * **level** (string, required): the redundancy level of the array (e.g. linear, raid1, raid5, etc.).
     * **devices** (list of strings, required): the list of devices (referenced by their absolute path) in the array.
     * **spares** (integer): the number of spares (if applicable) in the array.


ajeddeloh:
@bgilbert you mentioned wanting to document this. This is pretty minimal. Do you think we should document more (e.g. why we need to do this) and if so where; none of the existing docs seemed like a good place and I'm not sure it warrants its own section. There's https://coreos.com/ignition/docs/latest/boot-process.html but it seems much higher level than this.

bgilbert:
I don't think this belongs in the CT spec, since it's a feature of the OS and not of CLCs or Ignition configs per se.  I'd suggest creating a guide to root-on-RAID, probably as a section in a larger document about configuring storage.  Unfortunately there are no great candidates:
- [This](https://coreos.com/ignition/docs/latest/examples.html#create-a-raid-enabled-data-volume) is in the Ignition docs rather than the OS docs.
- [This](https://coreos.com/os/docs/latest/clc-examples.html#storage-and-files) is in the CLC docs and is really for introductory examples.
- [This](https://coreos.com/os/docs/latest/mounting-storage.html) is the right type of document but is scoped to data volumes.

Maybe this needs a new document scoped to configuring the root filesystem.  Root-on-LVM and root-on-LUKS could also go there once they work.

ajeddeloh:
I think that we should have a note here about it (although not go in depth why). CLCT is for CL specifcally, so I don't see an issue with adding a note about it. It is, in a way, part of the spec in that if you don't adhere to it you won't generate a valid (i.e. bootable) config. We make note that the `filesystem` field in the `files` section must correspond to filesytem in the `filesystems` section. Here we're stating something similar: that the root parition needs to be on a raid array suffixed with `-hasroot`

bgilbert:
I don't agree.  Your `files` example is a requirement imposed by Ignition, but the `-hasroot` suffix is essentially a configuration setting accepted by the OS.  There are plenty of documents on the site that detail how to configure the OS, using CLCs, in ways that can't be derived from first principles.  Most of that material is out of scope for the document that defines the semantics of a CLC.

Better example: if you're recreating the root filesystem, you must label it `ROOT` or CL won't boot.  We don't mention that in the spec either.

ajeddeloh:
I'd argue perhaps we should add a note about the `ROOT` label as well, but I don't really feel that strongly about it. Either way, what do you think about adding a section for special root partitions in the Mounting Storage docs?

bgilbert:
Basically none of the material in the introductory section of Mounting Storage applies to this case, which implies that it's not the right doc.  Downsides to a separate document?

ajeddeloh:
Seperate doc is probably the best option. Maybe put it after the mounting doc under the "setting up a cluster" docs?

bgilbert:
Or under "uncategorized docs" even.  We need a section for CL docs that aren't necessarily cluster-related, but we don't have that right now.

ajeddeloh:
SGTM, closing this.

PR #105, opened 2017-09-05
Produce binaries named "ct"

sj26:
I found myself really wanting a way to `go get` this package. It took my a while to figure out that the command is in the `internal` package. It seems conventional these days to put command packages in a `cmd/[name]` path. This way the following command installs the latest version of the `ct` command:

```
go get -u github.com/coreos/container-linux-config-transpiler/cmd/ct
```

🙇


crawford:
Doing this will expose `main.go` to being imported. Is it possible to keep this in the internal package (I haven't tested this myself)?

dgonyeo:
Go disallows importing packages with a `main` function. With these changes, this is the result of attempting to import `github.com/coreos/container-linux-config-transpiler/cmd/ct`:
```
main.go:5:2: import "github.com/coreos/container-linux-config-transpiler/cmd/ct" is a program, not an importable package
```

crawford:
What an unending maze of implicit rules...

I'd rather not introduce another top-level directory if we can help it. Will it work inside of internal?

sj26:
You could move it into `internal/ct` if you like. This should work:

```
go get -u github.com/coreos/container-linux-config-transpiler/internal/ct
```

`cmd/` is just a convention which might signal a runnable command, instead of `internal/` which signals "don't go in here".

Looking inside internal, it's just a version and a helper tool. Perhaps we could re-arrange slightly. Move the docs helper into `cmd/internal/docs/main.go`, and move the version constant into it's only consumer, `cmd/ct/`. I've attached some more commits, take a look?

sj26:
For your consideration I've also updated the `Dockerfile` to use a multi-stage process instead of needing to do multiple docker builds based on this work:

https://github.com/sj26/container-linux-config-transpiler/compare/name-binary...docker-multi-stage

crawford:
After discussing this further, I don't think this is the direction we want to go. `go get`, in its current state, has a number of problems and makes it difficult to help users debug. When installed with `go get`, CT will always report that it hasn't been correctly built (which is true). Thank you anyway.

dghubble:
Whoa, back from vacation and seeing this. The original request to be able to `go get` a `cmd/ct` is completely reasonable. Of course we want our command line tool to be go gettable.

I don't see much justification for the way this project is using `internal`. With Ignition `main.go` was hidden in internal out of some concern someone might try to run it on their local machine, and its unsafe, etc. Fine. But here, `ct` is a nice, but pretty ordinary YAML config to JSON command line utility. I see no legitimate reason for it to defy the conventions like this.

The flaw you describe is a reason we should improve the version printout.

@sj26 Let's please keep multi-stage Docker builds a separate discussion.

dghubble:
I suppose this is a PR - we can close this PR and say take a "different approach", but the original issue stands.

crawford:
The only way this works is if we hardcode the version string in the source. The moment we do this, it becomes worthless since the vast majority of builds will use a version like "X.X.X-dev", which doesn't actually point to a specific point. Hardcoding the string is also very error prone since it requires a human to keep it in sync (which we know from experience happens quite a bit).

@dghubble This versioning strategy is what we use for all of our Go binaries. If you have a solution which addresses my points above, I'd be willing to switch to that.

dghubble:
Folks using `go get -u github.com/coreos/container-linux-config-transpiler/cmd/ct` are aware this fetches and installs the "latest" version. They're aware its not an officially tagged release. That's the nature of the `-u` flag. In that light, "was not built properly" is a suitable version to print. If we really want something more descriptive, we could say "not an official build". Regardless of the printout, it doesn't carry with it all the trappings of an official release anyway. Seems like we're rationalizing that its not like an official release so we shouldn't allow it.

crawford:
> Seems like we're rationalizing that its not like an official release so we shouldn't allow it.

While I'm generally in favor of this stance, I'm actually arguing for something much more permissive: I don't want to allow CT to easily be built incorrectly. Reason being, I don't want to support CT if it has been built incorrectly. Users are inevitably going to report bugs with a version string of "was not built properly". Not only is this jarring for the users, it's frustrating for the people trying to support them.

dghubble:
People will always be able to clone your project and build an unofficial release that you don't consider "correct". Regardless of `go get`. This applies to any project. If you only wish to support official releases, you can say that and close the issue and/or make the version printout "unofficial, unsupported version".

This isn't a good reason to ignore go get functionality which should "just work" if we were structuring this right. Our tool is not special. It is not above the rules or conventions.

crawford:
> People will always be able to clone your project and build an unofficial release that you don't consider "correct". Regardless of go get. This applies to any project.

Yes, but it is much more difficult to incorrectly build the project this way. There is very clearly a Makefile, so `make` should be used. When using `go get`, there is no way to change the way a project is built. It seems more confusing to end-users that they would be able to `go get`, but the resulting binary cannot be supported.

> This isn't a good reason to ignore go get functionality which should "just work" if we were structuring this right. Our tool is not special. It is not above the rules or conventions.

I think this is another conversation, but I do think our tool is above the rules. I think _all_ binaries are above the rules since the rules are so poorly thought out. The world does not develop software in a monorepo; we need versions. That's exactly why http://labix.org/gopkg.in was built.

dghubble:
> I do think our tool is above the rules. I think all binaries are above the rules since the rules are so poorly thought out.

This doesn't seem like a productive way to built in the Go ecosystem and I can't really debate a dogma. To each their own I suppose.

PR #106, opened 2017-09-13
*: generate Ignition 2.1.0 configs

dgonyeo:
This pull request modifies ct to generate 2.1.0 configs instead of 2.0.0 configs. Container Linux Configs from before this change should still be valid, but may generate new warnings for deprecated fields.

The commits here:
- update the vendored ignition types package to be the 2.1 one
- make the necessary changes to build/run with the 2.1 package, producing 2.1.0 configs
- add in all the new fields available in the 2.1.0 spec
- update the tests to pass with these changes and add in tests to cover the new fields


config/types/files.go
```diff
@@ -45,80 +56,137 @@ type Remote struct {
...
-					User:       ignTypes.NodeUser{ID: &file.User.Id},
-					Group:      ignTypes.NodeGroup{ID: &file.Group.Id},
+	register2_0(convertFiles)
+}
+
+func convertFiles(in Config, ast validate.AstNode, out ignTypes.Config, platform string) (ignTypes.Config, report.Report, validate.AstNode) {
+	r := report.Report{}
+	files_node, _ := getNodeChildPath(ast, "storage", "files")
+	for i, file := range in.Storage.Files {
+		fmt.Printf("converting file %q\n", file.Path)
```
ajeddeloh: debugging left overs?

config/types/files.go
```diff
@@ -45,80 +56,137 @@ type Remote struct {
...
-		files_node, _ := getNodeChildPath(ast, "storage", "files")
-		for i, file := range in.Storage.Files {
-			file_node, _ := getNodeChild(files_node, i)
-			newFile := ignTypes.File{
-				Node: ignTypes.Node{
-					Filesystem: file.Filesystem,
-					Path:       file.Path,
-					User:       ignTypes.NodeUser{ID: &file.User.Id},
-					Group:      ignTypes.NodeGroup{ID: &file.Group.Id},
+	register2_0(convertFiles)
```
ajeddeloh: Any reason to break into a named function? I'm not against it, but maybe split into two seperate commit (break out into named fn, then the 2.1.0 changes) to make diff viewing a little easier.

config/types/files.go
```diff
@@ -70,123 +69,120 @@ type Link struct {
...
-		fmt.Printf("converting file %q\n", file.Path)
-		file_node, _ := getNodeChild(files_node, i)
-		newFile := ignTypes.File{
-			Node: ignTypes.Node{
-				Filesystem: file.Filesystem,
-				Path:       file.Path,
-				User: ignTypes.NodeUser{
-					ID:   file.User.Id,
-					Name: file.User.Name,
+	register2_0(func(in Config, ast validate.AstNode, out ignTypes.Config, platform string) (ignTypes.Config, report.Report, validate.AstNode) {
```
ajeddeloh: wasn't this changed in an earlier commit?

config/types/passwd.go
```diff
@@ -85,6 +84,8 @@ func init() {
 			}
 
 			if user.Create != nil {
+				fmt.Printf("user.Create != nil: %v\n", user.Create)
```
ajeddeloh: debugging left overs?

config/types/disks.go
```diff
@@ -106,5 +106,5 @@ func convertPartitionDimension(in string) (ignTypes.PartitionDimension, error) {
 	if b%BYTES_PER_SECTOR != 0 {
 		sectors++
 	}
-	return ignTypes.PartitionDimension(uint64(sectors)), nil
+	return int(sectors), nil
```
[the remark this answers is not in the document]
dgonyeo: That's a side effect of the json schema change. If this needs to be an int64, it'll need to be changed in Ignition, and that's autogenerated from json schema, so you might need to change (or make github issues for) the converter.

config/config_test.go
```diff
@@ -352,7 +352,7 @@ systemd:
 							Name:     "test1.service",
 							Enable:   true,
 							Contents: "test1 contents",
-							DropIns: []types.SystemdUnitDropIn{
```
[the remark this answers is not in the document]
dgonyeo: This is an artifact of the json schema change. It wasn't missed, 2.0.0 didn't use json schema. Since this PR is moving to 2.1, this PR is moving to the autogenerated json schema structs in Ignition, which results in naming changes like this.

config/types/files.go
dgonyeo: Oh I didn't intend for this to make it into the PR. Ignore this, I'll rip it out.

config/types/files.go
dgonyeo: yup, ignore this

config/types/files.go
dgonyeo: No I was trying to hunt down a problem. A later commit should undo this, I'll clean up the history

config/types/passwd.go
dgonyeo: yup

jbguerraz:
We can't wait to to get that in (we need the storage.links especially) :) 
Thx @dgonyeo !

dgonyeo:
@ajeddeloh history has been cleaned up, so should be ready for another pass

Also updated the vendor commit message.

dgonyeo:
still need to do the docs change

glide.lock
```diff
@@ -23,4 +23,16 @@ imports:
   - config/validate/report
 - name: github.com/vincent-petithory/dataurl
   version: 9a301d65acbb728fcc3ace14f45f511a4cfeea9c
-testImports: []
+testImports:
```
ajeddeloh: It doesn't look like we're actually pulling in the tests themselves (`glide-vc` is probably removing them). Is there a way to not need this?

dgonyeo: Added to the `ignore` section of `glide.yaml`, they shouldn't be included anymore.

config/types/passwd.go
ajeddeloh: Still present.

ajeddeloh:
We should add tests for new fields (e.g. `links`, `directories`, the new fields in `passwd`, etc)

config/types/passwd.go
dgonyeo: fixed

dgonyeo:
The changes to `config/config_test.go` in the final commit should test the new fields. Is that sufficient?

config/types/filesystems.go
```diff
@@ -21,9 +21,9 @@ import (
 )
 
 type Filesystem struct {
-	Name  string `yaml:"name"`
-	Mount *Mount `yaml:"mount"`
-	Path  string `yaml:"path"`
+	Name  string  `yaml:"name"`
```
ajeddeloh: Did this make its way into the wrong commit?

config/types/passwd.go
```diff
@@ -65,9 +65,6 @@ type Group struct {
 func init() {
 	register2_0(func(in Config, ast validate.AstNode, out ignTypes.Config, platform string) (ignTypes.Config, report.Report, validate.AstNode) {
 		for _, user := range in.Passwd.Users {
-			if user.Name != "core" && user.Create == nil {
```
ajeddeloh: Did this make its way into the wrong commit?

ajeddeloh:
Oh whoops, github collapsed that and I missed it.

config/types/filesystems.go
dgonyeo: Oh these were fixes for issues I found when writing the tests. I'll move them into the correct commit.

config/types/passwd.go
dgonyeo: Just moved it into a more appropriate commit

config/config_test.go
```diff
@@ -16,10 +16,13 @@ package config
 
 import (
 	"errors"
-	"reflect"
+	"net/url"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
```
ajeddeloh: Any reason for this change? Should be included in commit message.

config/config_test.go
```diff
@@ -534,12 +694,8 @@ flannel:
 
 	for i, test := range tests {
 		cfg, _, err := Parse([]byte(test.in.data))
-		if !reflect.DeepEqual(err, test.out.r) {
-			t.Errorf("#%d: bad error: want %v, got %v", i, test.out.r, err)
-		}
-		if !reflect.DeepEqual(cfg, test.out.cfg) {
-			t.Errorf("#%d: bad config: want %#v, got %#v", i, test.out.cfg, cfg)
-		}
+		assert.Equal(t, test.out.r, err, "#%d: bad report", i)
```
ajeddeloh: Does testify print the want/got messages?

dgonyeo: Yes, and it also pretty-prints a nice diff between the expected and actual structs, which was the motivation here for me to add it. Comparing the difference between walls of text is slow and hard when attempting to figure out issues.

config/config_test.go
dgonyeo: Updated commit message to mention this and rationale for using it.

config/config_test.go
dgonyeo: Turns out the packages I `ignore`'d earlier were necessary for `testify`, so I just readded them in a new commit for the `config/config_test.go` changes.

ajeddeloh:
Can you post the actual commit order (since github PR's list them in chronological order as opposed to the actual order)

dgonyeo:
Here's the output from `git log`, with all commits after `master`. Note the hashes will change as the review progresses.

```
commit 31b8b1798e62fcab33f1a9260a1ac9c56ec2ff04 (HEAD -> 2.1, dgonyeo/2.1)
Author: Derek Gonyeo <derek@gonyeo.com>
Date:   Mon Sep 18 14:30:05 2017 -0700

    vendor: add github.com/stretchr/testify/assert for config/config_test.go

commit 99dadbe39f574b36484108e45784b0fd9094a5a2
Author: Derek Gonyeo <derek@gonyeo.com>
Date:   Mon Sep 18 14:13:24 2017 -0700

    config/config_test.go: updates to work with 2.1 changes
    
    Also modifies the file to use github.com/stretchr/testify/assert for
    comparing two structs to be equal, so that differences are
    pretty-printed and thus far easier to debug.

commit 106a5b5522fea3f6b654e34327a63c9c86ce2945
Author: Derek Gonyeo <derek@gonyeo.com>
Date:   Mon Sep 11 15:35:18 2017 -0700

    config/types: add in new fields available in Ignition spec v2.1.0

commit be8a33ca284ce0ace34cb788f84dcdab38a77fd8
Author: Derek Gonyeo <derek@gonyeo.com>
Date:   Mon Sep 11 14:33:33 2017 -0700

    config/types/files: fix typo to invert logic for adding line/col info

commit 33cf0b6a7a904fa9e2d712b79a66d82e3f0dda95
Author: Derek Gonyeo <derek@gonyeo.com>
Date:   Tue Sep 12 13:03:57 2017 -0700

    vendor/github.com/ajeddeloh/yaml: update to get changes from upstream

commit cbc960adfc83e4a5b199c478a449494010fff32e
Author: Derek Gonyeo <derek@gonyeo.com>
Date:   Thu Sep 7 16:03:05 2017 -0700

    vendor: replace v2_0 with v2_1 ignition types

commit 56af9d438a13c9d0c5b46703143d5ee916255b51
Author: Derek Gonyeo <derek@gonyeo.com>
Date:   Thu Sep 7 16:03:27 2017 -0700

    config: changes to use Ignition v2_1 types
    
    Modifies the config package to use the Ignition v2_1 types package, with
    no added functionality or features.

```

ajeddeloh:
Should probably rebase to vendor, then add things that depend on the vendoring, but other than that and the `typeCasts.go` > `type_casts.go` fix, LGTM

dgonyeo:
Rebasing to vendor first doesn't really make sense with glide, since if you checked out that commit and ran `make vendor` it would undo the vendor changes (instead of do nothing) because at that point in the history `v2_0` would still be used and not `v2_1`

dgonyeo:
File rename complete

ajeddeloh:
Oh I guess that makes sense, in terms of vendoring after. It's not really due to glide but rather glide-vc.

dgonyeo:
Added docs

doc/configuration.md
```diff
@@ -57,12 +65,36 @@ _Note: all fields are optional unless otherwise marked_
...
     * **user** (object): specifies the file's owner.
       * **id** (integer): the user ID of the owner.
+      * **name** (string): the user name of the owner.
     * **group** (object): specifies the group of the owner.
       * **id** (integer): the group ID of the owner.
+      * **name** (string): the group name of the owner.
+  * **directories** (list of objects): the list of directories to be created.
+    * **filesystem** (string, required): the internal identifier of the filesystem in which to create the directory. This matches the last filesystem with the given identifier.
+    * **path** (string, required): the absolute path to the directory.
+    * **mode** (integer): the directory's permission mode. Note that the mode must be properly specified as a **decimal** value (i.e. 0755 -> 493).
```
ajeddeloh: Not necessary with yaml; you can specify numbers in octal. Drop the part about decimal value.

jbguerraz:
GG @dgonyeo ! thank you!
Any idea about when we could expect a new release with those changes ?

dgonyeo:
There's one other thing we'd like to get in before cutting a release, but it shouldn't take too long. Probably by the end of the week.

PR #107, opened 2017-09-15
config/types/locksmith: don't use custom types to validate fields

dgonyeo:
Change validation logic in the locksmith struct to attach all validation
functions to the top level locksmith struct, such that the string
special casing in config/types/common.go#getArgs can be relied on to
quote escape the strings, and %s can be used in the provided format
string instead of %q.

This bug was introduced in https://github.com/coreos/container-linux-config-transpiler/pull/66. It only affects the locksmith section because the format string the locksmith section provides to `getArgs` was `%q` to handle spaces before the PR that introduced the bug, and I believe that I didn't notice this bug when reviewing that PR because some of the fields in the `Locksmith` struct were type aliased so that validation functions could be attached, and thus failed the string special casing logic of `getArgs` that was added, and were formatted properly. Note that only some of the locksmith fields are double escaped.

Fixes https://github.com/coreos/bugs/issues/2155

### DIFF
--- a/config/types/locksmith.go
+++ b/config/types/locksmith.go
@@ -33,23 +33,19 @@ var (
 )
 
 type Locksmith struct {
-	RebootStrategy RebootStrategy `yaml:"reboot_strategy" locksmith:"REBOOT_STRATEGY"`
-	WindowStart    WindowStart    `yaml:"window_start"    locksmith:"LOCKSMITHD_REBOOT_WINDOW_START"`
-	WindowLength   WindowLength   `yaml:"window_length"   locksmith:"LOCKSMITHD_REBOOT_WINDOW_LENGTH"`
-	Group          string         `yaml:"group"           locksmith:"LOCKSMITHD_GROUP"`
-	EtcdEndpoints  string         `yaml:"etcd_endpoints"  locksmith:"LOCKSMITHD_ENDPOINT"`
-	EtcdCAFile     string         `yaml:"etcd_cafile"     locksmith:"LOCKSMITHD_ETCD_CAFILE"`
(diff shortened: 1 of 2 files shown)


config/types/update.go
```diff
@@ -74,6 +74,7 @@ func init() {
 			if len(lines) > 0 {
 				contents += "\n" + strings.Join(lines, "\n")
 			}
+			fmt.Println(lines)
```
arithx: Is this leftover debugging code?

config/types/locksmith.go
```diff
@@ -33,51 +33,47 @@ var (
...
 func (l Locksmith) Validate() report.Report {
 	if (l.WindowStart != "" && l.WindowLength == "") || (l.WindowStart == "" && l.WindowLength != "") {
 		return report.ReportFromError(ErrMissingStartOrLength, report.EntryError)
 	}
 	return report.Report{}
 }
 
-func (r RebootStrategy) Validate() report.Report {
-	switch strings.ToLower(string(r)) {
+func (l Locksmith) ValidateRebootStrategy() report.Report {
```
crawford: What ensures that this function is called now?

config/types/update.go
dgonyeo: Yup, removing

config/types/locksmith.go
dgonyeo: The validate code checks for functions on structs with `Validate<NAME>` where `<NAME>` is the name of the field, and calls them if they exist.

Here's the logic that does that: https://github.com/coreos/ignition/blob/58bcc85a20872686c2d6d2f8db2e439cd2e1ccb5/config/validate/validate.go#L185

crawford: Neat.

PR #108, opened 2017-09-15
config/types/locksmith: Do not double-quote string values

saj:
Fixes #2155.


saj:
Closing in preference for #107.

dgonyeo:
The changes need to be a little more complex than just changing the format string, since only _some_ of the fields are getting double escaped.

PR #109, opened 2017-09-20
Add aliases for type guids

ajeddeloh:
cc @dgonyeo

### DIFF
--- a/config/types/disks.go
+++ b/config/types/disks.go
@@ -27,6 +27,15 @@ const (
 	BYTES_PER_SECTOR = 512
 )
 
+var (
+	type_guid_map = map[string]string{
+		"raid_containing_root":  "be9067b9-ea49-4f15-b4f6-f36f8c9e1818",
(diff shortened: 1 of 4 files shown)


config/types/disks.go
```diff
@@ -44,6 +44,12 @@ type Partition struct {
 
 func init() {
 	register2_0(func(in Config, ast validate.AstNode, out ignTypes.Config, platform string) (ignTypes.Config, report.Report, validate.AstNode) {
+		type_guid_map := map[string]string{
```
dgonyeo: nit: I'd prefer this to go at the top of the file so it's easier to find

ajeddeloh:
Do you think it's worth mentioning what the aliases resolve to in the docs? The description in `configuration.md` should be short, so I don't want to include it there, but there doesn't seem to be a good place otherwise

ajeddeloh:
Added operator notes

PR #114, opened 2017-09-27
Change ConvertAs2_0 to ConvertAs2_1

ajeddeloh:
See https://github.com/coreos/bugs/issues/2171

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -62,13 +62,13 @@ func Parse(data []byte) (types.Config, astnode.AstNode, report.Report) {
 	return cfg, root, r
 }
 
-// ConvertAs2_0 will convert a golang struct representing a Container Linux
+// ConvertAs2_1 will convert a golang struct representing a Container Linux
 // Config into an Ignition Config, and a report of any warnings or errors. It
 // takes the parse tree from parsing the Container Linux config as well.
-// ConvertAs2_0 also accepts a platform string, which can either be one of the
+// ConvertAs2_1 also accepts a platform string, which can either be one of the
 // platform strings defined in config/templating/templating.go or an empty
 // string if [dynamic data](doc/dynamic-data.md) isn't used.
-func ConvertAs2_0(in types.Config, p string, ast astnode.AstNode) (ignTypes.Config, report.Report) {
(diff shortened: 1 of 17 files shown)


config/config.go
```diff
@@ -68,7 +68,7 @@ func Parse(data []byte) (types.Config, astnode.AstNode, report.Report) {
 // ConvertAs2_0 also accepts a platform string, which can either be one of the
```
ajeddeloh: Whoops, fixing this

ajeddeloh:
Missed a lot of thing, not done

ajeddeloh:
Ok, should have got them all now.

sdemos:
What is the point of renaming these functions? It seems like having the version numbers in the function names has no tangible benefits that I can see. Why shouldn't we just remove them and then never have to deal with this search and replace?

ajeddeloh:
I'd be fine with that. I think originally it wasn't clear if we'd support multiple output versions, so it was versioned, then when we bumped to 2.1.0, the function names were missed, leading to this PR.

@dgonyeo do you have a opinion on the matter?

sdemos:
I would personally be in favor of ripping the band-aid off and just dropping the version numbers altogether.

dghubble:
I assumed you'd be providing conversions to multiple Ignition spec versions, but I don't think that's the case. If the intent is to always produce a recent version (e.g. Ignition 2.1.0), I agree with @sdemos about dropping the numbers - vendoring the ct package provides the pinning.

dgonyeo:
I'm in favor of dropping the version numbers.

Unless we want to revisit the decision of ct only producing a recent version (as opposed to being able to produce multiple versions), there's no reason for these names to be tied to the Ignition schema version number.

sdemos:
@ajeddeloh do you have time to do that with this pr or should I take care of it and open a new one?

ajeddeloh:
I'm working on a grub/fat/something in very early boot problem right now, and wont have time for this until that's done (timeboxed on this friday). If you want to pick it up, feel free.

sdemos:
work to drop version numbers is in #127. closing this pr.

PR #115, opened 2017-09-28
Fix case of names in documentation to match YAML

errm:


ajeddeloh:
Good catch. Looks like we'll also need to fix the `httpReponseHeaders` and `httpTotal` in the `timeouts` section. 

cc @dgonyeo

errm:
Cool I have updated the `timeouts` section also now

ajeddeloh:
Thanks for getting the timeouts too.

PR #117, opened 2017-10-30
Update Dockerfile

leonkramer:
Dockerfile

Docker was unable to build container with current Dockerfile. 

With new Dockerfile build process does work. Also added -pretty parameter, as this is likely the output a user desires.


dgonyeo:
These changes look like it'll build the current master version of clct, instead of whatever currently exists on disk. Was this your intention?

leonkramer:
tbh, I did first assume that the Dockerfile should work without cloning the repo first.

I updated the pull request, so that 'make' is being used instead of 'build', as with a fresh clone of clct the Dockerfile does not work currently.

PR #118, opened 2017-11-01
Makefile: Add arm64 release

glevand:
Build arm64 binaries in `make release`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,11 +36,20 @@ vendor:
 clean:
 	$(Q)rm -rf bin
 
-.PHONY: release
-release: \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -36,12 +36,19 @@ vendor:
 clean:
 	$(Q)rm -rf bin
 
-.PHONY: release
-release: \
+release_aarch64 := \
```
crawford: This should be PHONY as well.

glevand:
Made `release_aarch64` and `release_x86_64` phony targets.

glevand:
Rebased to latest.

PR #120, opened 2017-11-20
config: enable coreos-metadata-sshkeys serice on openstack

sdemos:
when an openstack provider is specified to ct, a dropin is created to
override the provider value when coreos-metadata is run. this is because
there is no "openstack" provider, since there are two possible metadata
sources and we can't rely on either existing or discover which to use,
so it must be specified for us explicitly.

however, the dropin that was created failed to enable the ssh-keys
service. this would normally be done by ignition for providers where we
can retrieve ssh keys, but since openstack keys are only available if we
know which metadata service to use, we have to enable it in the dropin
instead of enabling it across the board for the openstack provider.

ref coreos/bugs#2247


sdemos:
rebased onto master

PR #110, opened 2017-09-20
README.md: replace 'make build' with 'make'

dgonyeo:
Prompted by https://github.com/coreos/container-linux-config-transpiler/pull/80#commitcomment-24316077

PR #122, opened 2017-12-01
readme: fix build instructions

ruuda:
There is no "build" target, but just `make` does produce a binary.


ruuda:
Oh, looks like there was #110 already. Feel free to close this.

lucab:
LGTM. I'm fine with taking this one over #110.

PR #124, opened 2017-12-01
Add support for local files

ruuda:
This adds one more file contents source, `local`, in addition to `inline` and `remote`. It takes a filename, and includes the contents of that file. This is useful when you don’t want to pollute the ign file with many large inline files, or when you don’t want to indent the contents of the file to fit the yaml syntax. Also, editing such a local config file is easier when it is a separate file, because an editor can enable format-specific features.

Local files are expected to be in the same directory as the ign file, so you could track both under source control; this is just for convenience.

Currently this uses the path as is, so local paths are interpreted relative to the working directory. It would be nice to have them relative to the ign file, so the output does not depend on where `ct` is invoked. However, this is not always possible (when the input is stdin), and I am not sure what the best way to thread the input filename through the program is, for the case where it is available. Feedback would be appreciated.

### DIFF
--- a/config/types/files.go
+++ b/config/types/files.go
@@ -15,6 +15,7 @@
 package types
 
 import (
+	"io/ioutil"
 	"net/url"
 
 	"github.com/coreos/container-linux-config-transpiler/config/astyaml"
@@ -47,6 +48,7 @@ type File struct {
 type FileContents struct {
 	Remote Remote `yaml:"remote"`
 	Inline string `yaml:"inline"`
+	Local  string `yaml:"local"`
 }
 
 type Remote struct {
@@ -101,6 +103,28 @@ func init() {
 					Source: (&url.URL{
 						Scheme: "data",
 						Opaque: "," + dataurl.EscapeString(file.Contents.Inline),
+					}).String(),
+				}
+			}
+
+			if file.Contents.Local != "" {
+				contents, err := ioutil.ReadFile(file.Contents.Local)
+				if err != nil {
+					// If the file could not be read, record error and continue.
+					convertReport := report.ReportFromError(err, report.EntryError)
+					if n, err := getNodeChildPath(file_node, "contents", "local"); err == nil {
+						line, col, _ := n.ValueLineCol(nil)
+						convertReport.AddPosition(line, col, "")
+					}
+					r.Merge(convertReport)
+					continue
+				}
+
+				// Include the contents of the local file as if it were provided inline.
+				newFile.Contents = ignTypes.FileContents{
+					Source: (&url.URL{
+						Scheme: "data",
+						Opaque: "," + dataurl.Escape(contents),
 					}).String(),
 				}
 			}
--- a/config/types/files.go
+++ b/config/types/files.go
@@ -15,8 +15,11 @@
 package types
 
 import (
+	"errors"
+	"flag"
 	"io/ioutil"
 	"net/url"
+	"path"
 
 	"github.com/coreos/container-linux-config-transpiler/config/astyaml"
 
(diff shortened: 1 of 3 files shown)


doc/configuration.md
```diff
@@ -55,6 +55,7 @@ _Note: all fields are optional unless otherwise marked_
     * **path** (string, required): the absolute path to the file.
     * **contents** (object): options related to the contents of the file.
       * **inline** (string): the contents of the file.
+      * **local** (string): the path to a local file, on the machine where `ct` runs.
```
ajeddeloh: "the file on the machine where `ct` is run containing the contents of the file to be created"

cc @dgonyeo Do you have any idea for better wording?

ajeddeloh:
Thanks for the PR (and the other one fixing the report merging bug)! Tested and ran this; it works great. 

Minor nit: commit title should be `config/types/files` not `config/files/types`.

cc @dgonyeo @lucab

ruuda:
> commit title should be `config/types/files` not `config/files/types`

Fixed.

ruuda:
Is there anything I can do to move this forward? Should I update the wording of the documentation, or does anybody have a better suggestion still?

doc/configuration.md
sdemos: I think "the relative path to a file on the machine where `ct` is run. The contents of the file are included in the generated config." might be sufficient.

dgonyeo:
Thanks for the contribution! The code and docs look fine (I'll give them a more thorough read through tomorrow), and this feature would definitely improve ct's usability, but I want to spend a little more time thinking through the implications of this change. Adding a dependency on files on disk is a big enough change to how ct works that I want to make sure it isn't going to back us into a corner down the road. I'll talk it over with some of the others tomorrow and should hopefully have an update on this by EOD tomorrow.

dgonyeo:
After a little discussion we'd like to see this merged, but with a couple things changed:

- A `--files-dir` flag that specifies which directory to look for files in, instead of looking in the current directory ct was run in
- A [`Validate`](https://github.com/coreos/container-linux-config-transpiler/blob/178372799c80d86eb20cbe373d4519f2bcfc5d7f/config/types/etcd.go#L51) function that checks that the `--files-dir` flag was set if a local file is specified. The validation logic ct uses comes from Ignition so unfortunately plumbing the value down into the validation logic is more involved than makes sense for this, so you'll probably want to do something like use [`flag.Lookup`](https://golang.org/pkg/flag/#FlagSet.Lookup) to check for the flag.

We think ct's behavior will be more predictable with this flag by making it more obvious when there is a dependency on external files.

It would be great if you've got the spare time to make these changes, if you don't I can take over this PR.

ruuda:
A `--files-dir` flag seems reasonable to me, that sidesteps the stdin problem neatly, and I agree that it is more predictable.

Thanks for the pointers! I think I could find some time to work on this later this month. If you want to work on it earlier feel free to do so.

ruuda:
I added a `--files-dir` flag and updated local file support to read from there, please take another look.

doc/configuration.md
ajeddeloh: This should be updated to indicate that `--files-dir` is required if it is used.

config/types/files.go
```diff
@@ -108,7 +111,21 @@ func init() {
 			}
 
 			if file.Contents.Local != "" {
-				contents, err := ioutil.ReadFile(file.Contents.Local)
+				// The provided local file path is relative to the value of the
+				// --files-dir flag.
+				filesDir := flag.Lookup("files-dir")
+				if filesDir == nil || filesDir.Value.String() == "" {
+					err := errors.New("Local files require setting the --files-dir flag to the directory that contains the file.")
```
dgonyeo: nit: other errors typically don't have sentence formatting. Can you lowercase `Local` and drop the period at the end?

ruuda: Fixed.

ruuda:
> Please update the commit message on `2b4bd629a7445b29a17a0e83481ab83bb66003c2` to indicate the files should be in the directory specified by `--files-dir` instead of the directory with the ignition file.

Fixed. Kudos for caring about the commit history. I also rebased on top of master.

> This should be updated to indicate that `--files-dir` is required if it is used.

Fixed. I also clarified that local files get included in the config but remote files are not.

PR #126, opened 2017-12-12
doc/example.md: include more methods to generate password hashes

dgonyeo:
Pulls in block of ways to generate password hashes from [cloudinit's
documentation][1].

[1]: https://github.com/coreos/coreos-cloudinit/blame/master/Documentation/cloud-config.md#L415-L434

### DIFF
--- a/doc/examples.md
+++ b/doc/examples.md
@@ -45,21 +45,27 @@ passwd:
       shell: /bin/bash
 ```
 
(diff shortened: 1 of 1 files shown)


dgonyeo:
Also updates the example description on line 48 to match the example (the shell was changed in the example a bit ago).

dgonyeo:
The python example was updated for python 3, to prompt the user for the password, and to generate a random salt. SO didn't provide me with an improved version of the perl command, and I lack the perl skills to quickly update it with the same improvements.

PR #127, opened 2017-12-13
*: remove versioning from function names

sdemos:
for a while there was a thought that ct would provide the functionality
to convert between mutliple different specs, hence having versioned
functions for each different spec version. this never manifested itself,
and the effect was that everytime the spec bumped, we would have to
search and replace every function name with the new spec version,
constantly breaking api compatibility for no particular benefit.
removing the version numbers fixes this problem.

/cc @ajeddeloh @dgonyeo @dghubble


config/config_test.go
```diff
@@ -1611,13 +1611,13 @@ func TestConvertAs2_0(t *testing.T) {
...
 	for i, test := range tests {
-		cfg, r := ConvertAs2_0(test.in.cfg, "", nil)
+		cfg, r := Convert(test.in.cfg, "", nil)
 		assert.Equal(t, test.out.r, r, "#%d: bad report", i)
 		assert.Equal(t, test.out.cfg, cfg, "#%d: bad config", i)
 	}
 }
 
-func TestAndConvertAs2_0_0(t *testing.T) {
+func TestAndConvert_0(t *testing.T) {
```
ajeddeloh: I thinkk this is supposed to be TestParseAndConvert

ajeddeloh:
Other than the `ParseAndConvert` change, LGTM. I assume you've grepped for 2_0?

config/config_test.go
sdemos: oh yup, :s/ messed that one up.

sdemos: should I add parse there? it wasn't there before. I'm fine adding it if that makes sense.

sdemos:
`grep -ir "[[:digit:]]_[[:digit:]]" .` output tells me I didn't miss anything

config/config_test.go
ajeddeloh: Yeah. I added that test; it's supposed to be TestParseAndConvert. Looks like I just typo'd it.

sdemos:
my understanding of the matchbox impact is that it will be minimal. the dependency is pinned anyway, and when he goes through and updates the dependencies at some point in the future he will just have to change some function calls. I'm going to go ahead and merge this.

PR #128, opened 2017-12-23
readme: remove '--branch' from git clone

AlexxNica:
The branch previously specified (v0.3.1) doesn't seem to exist.


dgonyeo:
The branch flag works because it'll find the 0.3.1 tag. This tag is rather old however, would you mind reopening this to change it to v0.5.0?

PR #129, opened 2017-12-23
readme: update git clone --branch from v0.3.1 to v0.5.0

AlexxNica:
the branch previously specified (v0.3.1) doesn't seem to exist.

UPDATE: `v0.3.1` was actually based on a tag that I didn't see but happened to be outdated, `v0.5.0` is the latest one, so a new commit was made to reflect that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ The easiest way to get started using ct is to download one of the binaries from 
 To build from source you'll need to have the go compiler installed on your system.
 
 ```shell
-git clone https://github.com/coreos/container-linux-config-transpiler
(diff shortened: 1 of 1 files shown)


dgonyeo:
Whoops, was going through my email and saw the other closed version of this first.

The `--branch` flag works for me because it finds the `v0.3.1` tag. We specify this because we want people to be using tagged releases, not whatever is currently on master.

The `v0.3.1` tag is rather old though, would you mind changing this to update that line to `v0.5.0`?

AlexxNica:
@dgonyeo Oops, sorry about that! I've already updated the PR, take a look when you get a chance! 😄

PR #134, opened 2018-01-04
config/types: default unset modes to sensible values

dgonyeo:
When the mode for a file or directory is unset in a CLC, default them to
0644 and 0755 respectively, and emit a warning.

Based on a pre-holiday conversation with @ajeddeloh and @bgilbert

### DIFF
--- a/config/types/files.go
+++ b/config/types/files.go
@@ -17,16 +17,26 @@ package types
 import (
 	"errors"
 	"flag"
+	"fmt"
 	"io/ioutil"
 	"net/url"
 	"path"
 
 	"github.com/coreos/container-linux-config-transpiler/config/astyaml"
+	"github.com/coreos/container-linux-config-transpiler/internal/util"
 
 	ignTypes "github.com/coreos/ignition/config/v2_1/types"
 	"github.com/coreos/ignition/config/validate/astnode"
 	"github.com/coreos/ignition/config/validate/report"
 	"github.com/vincent-petithory/dataurl"
+)
+
+var (
+	DefaultFileMode = 0644
+	DefaultDirMode  = 0755
+
+	WarningUnsetFileMode = fmt.Errorf("mode unspecified for file, defaulting to %#o", DefaultFileMode)
+	WarningUnsetDirMode  = fmt.Errorf("mode unspecified for directory, defaulting to %#o", DefaultDirMode)
 )
 
 type FileUser struct {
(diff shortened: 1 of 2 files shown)


dgonyeo:
Updated to fix the tests

config/types/files.go
```diff
@@ -77,11 +78,22 @@ type Link struct {
 	Target     string    `yaml:"target"`
 }
 
+func intToPtr(i int) *int {
```
ajeddeloh: Do we want to break this out into a util library?

dgonyeo: Honestly I find myself implementing so much dumb stuff like this all over the place I'm tempted to make a "missing from stdlib" library for us.

But until I snap and do that, sure.

dgonyeo: Updated

lucab:
@dgonyeo does this handle `mode: 0000`?

ajeddeloh:
@lucab haven't tested it, but it should. `0000` should be deserialize as `0` while omitting it should deserialize as `nil`

dgonyeo:
@lucab just tested, `mode: 0000` does not produce any warnings and doesn't get changed to the default.

config/types/files.go
```diff
@@ -82,6 +84,13 @@ func init() {
 		r := report.Report{}
 		files_node, _ := getNodeChildPath(ast, "storage", "files")
 		for i, file := range in.Storage.Files {
+			if file.Mode == nil {
```
ajeddeloh: why is this not in the validation logic?

dgonyeo: Moved the report logic to validation functions on `File` and `Directory`

config/types/files.go
```diff
@@ -77,11 +79,36 @@ type Link struct {
...
+		})
+	}
+	return r
+}
+
+func (d Directory) ValidateMode() report.Report {
+	r := report.Report{}
+	if d.Mode == nil {
+		r.Add(report.Entry{
+			Message: fmt.Sprintf("mode unspecified for directory %q, defaulting to 0644", d.Path),
```
ajeddeloh: 0755

dgonyeo: Ah yup. Fixed.

config/types/files.go
```diff
@@ -187,6 +214,9 @@ func init() {
 			out.Storage.Files = append(out.Storage.Files, newFile)
 		}
 		for _, dir := range in.Storage.Directories {
+			if dir.Mode == nil {
+				dir.Mode = util.IntToPtr(0755)
```
bgilbert: How about making these named constants, and then using the constants when building the `report.Entry` message?

dgonyeo: done

ajeddeloh: I think he meant making the default modes constants and building the error message from them.

dgonyeo: Updated. Does that look like overkill?

ajeddeloh: You should use the constant here as well

dgonyeo: whoops, yeah. Done.

PR #135, opened 2018-01-12
news: add news for v0.6.0

dgonyeo:


NEWS
```diff
@@ -1,3 +1,23 @@
+25-Sep-2017 CONFIG-TRANSPILER v0.6.0
```
crawford: Date isn't correct.

dgonyeo: Fixed

PR #138, opened 2018-01-30
Documentation: how to download and verify signature

czerasz:


README.md
```diff
@@ -30,6 +30,27 @@ There are plenty of small, self-contained examples [in the documentation][exampl
...
+# Specify download URL
+DOWNLOAD_URL=https://github.com/coreos/container-linux-config-transpiler/releases/download
+
+# Download Config Transpiler binary
+curl -L ${DOWNLOAD_URL}/${CT_VER}/ct-${CT_VER}-x86_64-unknown-linux-gnu -o /tmp/ct-${CT_VER}-x86_64-unknown-linux-gnu
+chmod u+x /tmp/ct-${CT_VER}-x86_64-unknown-linux-gnu
+
+# Download and import CoreOS application signing GPG key
+curl https://coreos.com/dist/pubkeys/app-signing-pubkey.gpg -o /tmp/app-signing-pubkey.gpg
+gpg --import --keyid-format LONG /tmp/app-signing-pubkey.gpg
```
ajeddeloh: Should use `gpg2` since `gpg` can be either version 1 or 2 depending on distro.

README.md
```diff
@@ -30,6 +30,27 @@ There are plenty of small, self-contained examples [in the documentation][exampl
...
+One can use the following script to download and verify the signature of Config Transpiler:
+
+```bash
+# Sepcify Config Transpiler version
+CT_VER=v0.6.1
+# Specify download URL
+DOWNLOAD_URL=https://github.com/coreos/container-linux-config-transpiler/releases/download
+
+# Download Config Transpiler binary
+curl -L ${DOWNLOAD_URL}/${CT_VER}/ct-${CT_VER}-x86_64-unknown-linux-gnu -o /tmp/ct-${CT_VER}-x86_64-unknown-linux-gnu
```
ajeddeloh: AFAICT this should also work on macOS, maybe add `${OS}` like `CT_VER` and `DOWNLOAD_URL` and list the options as a comment? I don't have a mac to test this however.

czerasz:
@ajeddeloh I added Your feedback. Good suggestions!

README.md
```diff
@@ -30,6 +30,33 @@ There are plenty of small, self-contained examples [in the documentation][exampl
...
+DOWNLOAD_URL=https://github.com/coreos/container-linux-config-transpiler/releases/download
+
+# Remove previous downloads
+rm -f /tmp/ct-${CT_VER}-x86_64-${OS} /tmp/ct-${CT_VER}-x86_64-${OS}.asc /tmp/coreos-app-signing-pubkey.gpg
+
+# Download Config Transpiler binary
+curl -L ${DOWNLOAD_URL}/${CT_VER}/ct-${CT_VER}-x86_64-${OS} -o /tmp/ct-${CT_VER}-x86_64-${OS}
+chmod u+x /tmp/ct-${CT_VER}-x86_64-${OS}
+
+# Download and import CoreOS application signing GPG key
```
ajeddeloh: Maybe change this to say "Download and import CoreOS application signing GPG key if it has not already been imported"

README.md
```diff
@@ -30,6 +30,33 @@ There are plenty of small, self-contained examples [in the documentation][exampl
...
+# Sepcify Config Transpiler version
+CT_VER=v0.6.1
+# Sepcify OS
+OS=apple-darwin # MacOS
+OS=unknown-linux-gnu # Linux
+# Specify download URL
+DOWNLOAD_URL=https://github.com/coreos/container-linux-config-transpiler/releases/download
+
+# Remove previous downloads
+rm -f /tmp/ct-${CT_VER}-x86_64-${OS} /tmp/ct-${CT_VER}-x86_64-${OS}.asc /tmp/coreos-app-signing-pubkey.gpg
```
ajeddeloh: I'd be in favor of downloading to temporary files then using `mv` to overwrite the old ones if everything downloaded and verified ok. That would prevent leaving users with nothing if something fails later. Either that change `-f` to `-i` so users don't accidently delete things they didn't want to.

czerasz: I used the same strategy as here:

screenshot from 2018-02-05 21-22-23

`/tmp` is a temporary directory. Stuff inside this directory doesn't survive the reboot. I wouldn't bother about removing files which are critical for the user.
The `-f` flag is used to not throw an error if the file is not available.

What do You think @ajeddeloh ?

README.md
```diff
@@ -30,6 +30,33 @@ There are plenty of small, self-contained examples [in the documentation][exampl
 
 The easiest way to get started using ct is to download one of the binaries from the [releases page on GitHub][releases].
 
+One can use the following script to download and verify the signature of Config Transpiler:
+
+```bash
+# Sepcify Config Transpiler version
```
dgonyeo: nit: `Specify`

README.md
```diff
@@ -30,6 +30,33 @@ There are plenty of small, self-contained examples [in the documentation][exampl
...
+CT_VER=v0.6.1
+# Sepcify OS
```
dgonyeo: nit: `Specify`

README.md
```diff
@@ -30,6 +30,33 @@ There are plenty of small, self-contained examples [in the documentation][exampl
...
+OS=apple-darwin # MacOS
+OS=unknown-linux-gnu # Linux
```
dgonyeo: For users copy/pasting/running this script, setting one OS and then immediately overriding it with another is redundant. Perhaps default to `OS=unknown-linux-gnu`, and having commented out lines above it for MacOS and windows?

README.md
```diff
@@ -30,6 +30,33 @@ There are plenty of small, self-contained examples [in the documentation][exampl
...
+# Specify download URL
+DOWNLOAD_URL=https://github.com/coreos/container-linux-config-transpiler/releases/download
+
+# Remove previous downloads
+rm -f /tmp/ct-${CT_VER}-x86_64-${OS} /tmp/ct-${CT_VER}-x86_64-${OS}.asc /tmp/coreos-app-signing-pubkey.gpg
+
+# Download Config Transpiler binary
+curl -L ${DOWNLOAD_URL}/${CT_VER}/ct-${CT_VER}-x86_64-${OS} -o /tmp/ct-${CT_VER}-x86_64-${OS}
```
dgonyeo: It would be nice if the arch was also a variable, so users could set `ARCH=aarch64` or something like that

README.md
```diff
@@ -30,6 +30,40 @@ There are plenty of small, self-contained examples [in the documentation][exampl
...
+
+# Specify download URL
+DOWNLOAD_URL=https://github.com/coreos/container-linux-config-transpiler/releases/download
+
+# Remove previous downloads
+rm -f /tmp/ct-${CT_VER}-${ARCH}-${OS} /tmp/ct-${CT_VER}-${ARCH}-${OS}.asc /tmp/coreos-app-signing-pubkey.gpg
+
+# Download Config Transpiler binary
+curl -L ${DOWNLOAD_URL}/${CT_VER}/ct-${CT_VER}-${ARCH}-${OS} -o /tmp/ct-${CT_VER}-${ARCH}-${OS}
+chmod u+x /tmp/ct-${CT_VER}-${ARCH}-${OS}
```
dghubble: Doesn't seem needed?

dgonyeo: Presumably a user will want to execute ct after they've verified that their copy is signed properly. Are you saying this step is implied and thus we should omit it?

dghubble: Seems fine, though not quite related to doing the gpg verify.

czerasz: To be able to execute `ct` you need to `chmod u+x` it.
The proper rights are missing on the downloaded file:

```bash
$ ls -l /tmp/ct-v0.6.1-x86_64-unknown-linux-gnu                                                 
-rw-r--r-- 1 czerasz czerasz 6158287 Feb 14 13:05 /tmp/ct-v0.6.1-x86_64-unknown-linux-gnu
```

README.md
ajeddeloh: oh, yeah hadn't thought about it being /tmp SGTM

README.md
```diff
@@ -30,6 +30,40 @@ There are plenty of small, self-contained examples [in the documentation][exampl
...
+
+# Download and import CoreOS application signing GPG key
+curl https://coreos.com/dist/pubkeys/app-signing-pubkey.gpg -o /tmp/coreos-app-signing-pubkey.gpg
+gpg --import --keyid-format LONG /tmp/coreos-app-signing-pubkey.gpg
```
ajeddeloh: Missed a `gpg2`, otherwise LGTM.

PR #139, opened 2018-01-31
Support custom platforms

ajeddeloh:
Add the ability to bring your own metadata service. Add a "custom" platform type and document how to use it by overwriting parts of the coreos-metadata service

### DIFF
--- a/config/templating/templating.go
+++ b/config/templating/templating.go
@@ -76,6 +76,13 @@ var platformTemplatingMap = map[string]map[string]string{
 	platform.CloudStackConfigDrive: {
 		fieldHostname: "CLOUDSTACK_LOCAL_HOSTNAME",
 	},
+	platform.Custom: {
+		fieldHostname:  "COREOS_CUSTOM_HOSTNAME",
+		fieldV4Private: "COREOS_CUSTOM_PRIVATE_IPV4",
+		fieldV4Public:  "COREOS_CUSTOM_PUBLIC_IPV4",
+		fieldV6Private: "COREOS_CUSTOM_PRIVATE_IPV6",
+		fieldV6Public:  "COREOS_CUSTOM_PUBLIC_IPV6",
+	},
 }
 
 // HasTemplating returns whether or not any of the environment variables present
--- a/doc/dynamic-data.md
+++ b/doc/dynamic-data.md
@@ -24,11 +24,56 @@ This is the information available in each provider.
 | OpenStack-Metadata | ✓          | ✓              | ✓             |                |               |
 | Vagrant-Virtualbox | ✓          | ✓              |               |                |               |
 
+## Custom metadata providers
+
+`ct` also supports custom metadata providers. To use the `custom` platform, modify the coreos-metadata service unit to execute your own custom metadata fetcher. The custom metadata fetcher must write an environment file `/run/metadata/coreos` defining a `COREOS_CUSTOM_*` environment variable for every piece of dynamic data used in the specified Container Linux Config. The environment variables are the same as in the Container Linux Config, but prefixed with `COREOS_CUSTOM_`.
+
+### Example
+
+Assume `https://example.com/metadata-script.sh` is a script which communicates with a metadata service and then writes the following file to `/run/metadata/coreos`:
+```
+COREOS_CUSTOM_HOSTNAME=foobar
+COREOS_CUSTOM_PRIVATE_IPV4=<The instance's private ipv4 address>
+COREOS_CUSTOM_PUBLIC_IPV4=<The instance's public ipv4 address>
+```
+
+The following Container Linux Config downloads the metadata fetching script, replaces the ExecStart line in `coreos-metadata` service to use the script instead, and configures etcd using the metadata provided. Use the `--platform=custom` flag when transpiling.
+```yaml container-linux-config
+storage:
+  files:
+    - filesystem: "root"
+      path: "/opt/get-metadata.sh"
+      mode: 0755
+      contents:
+        remote:
+          url: "https://example.com/metadata-script.sh"
+
+systemd:
+  units:
+    - name: "coreos-metadata.service"
+      dropins:
+       - name: "use-script.conf"
+         contents: |
+           [Service]
+           # Empty ExecStart= prevents the previously defined ExecStart from running
+           ExecStart=
+           ExecStart=/opt/get-metadata.sh
+
+etcd:
+  version:                     "3.0.15"
+  name:                        "{HOSTNAME}"
+  advertise_client_urls:       "http://{PRIVATE_IPV4}:2379"
+  initial_advertise_peer_urls: "http://{PRIVATE_IPV4}:2380"
+  listen_client_urls:          "http://0.0.0.0:2379"
+  listen_peer_urls:            "http://{PRIVATE_IPV4}:2380"
+  initial_cluster:             "{HOSTNAME}=http://{PRIVATE_IPV4}:2380"
+```
+
 ## Behind the scenes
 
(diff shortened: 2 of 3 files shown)


doc/dynamic-data.md
```diff
@@ -24,6 +24,49 @@ This is the information available in each provider.
 | OpenStack-Metadata | ✓          | ✓              | ✓             |                |               |
 | Vagrant-Virtualbox | ✓          | ✓              |               |                |               |
 
+## Custom metadata providers
+
+As of ct 7.0(FIXME: what do we put here), ct also supports custom metadata providers. To use the `custom` platform, overwrite the coreos-metadata service unit to execute your own custom metadata fetcher. The custom metadata fetcher much write an enviroment file with all of the dynamic fields used to `/run/metadata/coreos`.
```
dgonyeo: imo we don't need to say what version this exists in, just release a new version with this quick-ish and if people ask we can show them the NEWS file.

bgilbert: s/much/must/
s/enviroment/environment/
`dynamic fields used to /run/metadata/coreos`?

doc/dynamic-data.md
```diff
@@ -24,6 +24,49 @@ This is the information available in each provider.
...
+PUBLIC_IPV4=<The instance's public ipv4 address>
+```
+
+The following Container Linux Config downloads the metadata fetching script, replaces the ExecStart line in `coreos-metadata` service to use the script instead, and configures etcd using the metadata provided.
+```yaml
+storage:
+  files:
+    - filesystem: "root"
+      path: "/opt/get-metadata.sh"
+      mode: 0766
```
bgilbert: World-writable?

dgonyeo:
The shell script could even be sourced from the local disk at conversion time with the recent local files feature that was added

doc/dynamic-data.md
```diff
@@ -24,6 +24,49 @@ This is the information available in each provider.
...
+          url: "https://example.com/metadata-script.sh"
+
+systemd:
+  units:
+    - name: "coreos-metadata.service"
+      dropins:
+       - name: "use-script.conf"
+         contents: |
+           [Service]
+           ExecStart=
```
bgilbert: Needs a comment.

bgilbert:
As discussed OOB, we're going to leave out the sugar for now to avoid creating a compatibility constraint, but may add it later once we have more experience with the feature.

ajeddeloh:
Addressed various typos and fat-fingerings of file modes. Added some comments and made the language a little clearer.

config/templating/templating.go
```diff
@@ -76,6 +76,13 @@ var platformTemplatingMap = map[string]map[string]string{
 	platform.CloudStackConfigDrive: {
 		fieldHostname: "CLOUDSTACK_LOCAL_HOSTNAME",
 	},
+	platform.Custom: {
+		fieldHostname:  "HOSTNAME",
```
bgilbert: Should the variable names be namespaced like the others?

ajeddeloh: I was unsure. From an implementation standpoint, it makes a lot of sense. From a user I think it makes more sense to use what they are using in their templated config so they don't have to mental map them. It's just a little bit more they need to understand (although it is documented below). Either way I don't feel strongly.

ajeddeloh:
Updated to namedspace everything with `COREOS_` and updated the docs to match.

doc/dynamic-data.md
```diff
@@ -24,6 +24,51 @@ This is the information available in each provider.
 | OpenStack-Metadata | ✓          | ✓              | ✓             |                |               |
 | Vagrant-Virtualbox | ✓          | ✓              |               |                |               |
 
+## Custom metadata providers
+
+`ct` also supports custom metadata providers. To use the `custom` platform, modify the coreos-metadata service unit to execute your own custom metadata fetcher. The custom metadata fetcher must write an environment file to `/run/metadata/coreos` which defines all the above environment variables used in the Container Linux Config. The environment variables are the same as in the Container Linux Config, but prefixed with `COREOS_`.
```
bgilbert: `COREOS_CUSTOM_` now.

What happens if only some variables are defined?  There might not be all four of public/private IPv4/IPv6.

ajeddeloh: Oops, will fix that.

`ct` has no way of knowing what the custom metadata service will be able to provide, so it always allows them. I just tested and systemd will resolve an undefined environment variable to an empty string.

ajeddeloh: My intent with that sentence was to say that everything the user uses in their CLC must written out by the custom metadata service. What about:

"The custom metadata fetcher must write an environment file to `/run/metadata/coreos` defining every `COREOS_CUSTOM_*` environment variable used in the specified Container Linux Config"

ajeddeloh: Changed to that.

doc/dynamic-data.md
```diff
@@ -24,6 +24,51 @@ This is the information available in each provider.
 | OpenStack-Metadata | ✓          | ✓              | ✓             |                |               |
 | Vagrant-Virtualbox | ✓          | ✓              |               |                |               |
 
+## Custom metadata providers
+
+`ct` also supports custom metadata providers. To use the `custom` platform, modify the coreos-metadata service unit to execute your own custom metadata fetcher. The custom metadata fetcher must write and environment file `/run/metadata/coreos` defining a `COREOS_CUSTOM_*` environment variable for every piece of dynamic data used in the specified Container Linux Config. The environment variables are the same as in the Container Linux Config, but prefixed with `COREOS_CUSTOM_`.
```
bgilbert: `write an`

doc/dynamic-data.md
```diff
@@ -24,6 +24,51 @@ This is the information available in each provider.
...
+
+Example:
```
bgilbert:
```
### Example
```

doc/dynamic-data.md
```diff
@@ -24,6 +24,51 @@ This is the information available in each provider.
...
+
+This example assumes `https://example.com/metadata-script.sh` is a script which communicates with a metadata service then writes the following file to `/run/metadata/coreos`:
```
bgilbert: s/This example assumes/Assume/
s/then/and then/

doc/dynamic-data.md
```diff
@@ -24,6 +24,51 @@ This is the information available in each provider.
...
+```
+COREOS_CUSTOM_HOSTNAME=foobar
+COREOS_CUSTOM_PRIVATE_IPV4=<The instance's private ipv4 address>
+COREOS_CUSTOM_PUBLIC_IPV4=<The instance's public ipv4 address>
+```
+
+The following Container Linux Config downloads the metadata fetching script, replaces the ExecStart line in `coreos-metadata` service to use the script instead, and configures etcd using the metadata provided. Use the `--platform=custom` flag when transpiling.
+```yaml
```
bgilbert: Should this be `yaml container-linux-config:custom`?

ajeddeloh:
Updated the grammar. Left off the `:custom` for the config type until the next CT release happens (since the docs sync doesn't use master IIRC)

dghubble:
If I understand this correctly, this is targeting a user who:

* writes a custom metadata fetching script
* Adds that script on the filesystem
* Adds a coreos-metadata.service unit that uses that script

all so that the user can write `http://{PRIVATE_IPV4}:2379` instead of `http://{COREOS_CUSTOM_PRIVATE_IPV4}:2379` in the CLC. Was this a problem folks were having?

ajeddeloh:
Yeah, there was a coreos-user thread about it. CT doesn't allow any `{ }` templating without specifying a platform, regardless of what the value inside the `{}` is. This also makes it so the same CLC can target baremetal and non-baremetal platforms.

PR #140, opened 2018-02-02
travis: Build & push alpine docker image to quay if on master

stuart-warren:
rebased to master and changed to alpine to have more use with ci systems like gitlab-ci


stuart-warren:
ping @sdemos

sdemos:
thanks for updating this pr! sorry I haven't responded in a timely manner, things have been pretty busy here for the last couple weeks.

can you go into more detail about the alpine change in this pr? you mention that it's more useful with ci systems but I don't think I understand why that would be the case. I would prefer to keep the container as small as possible, but if there is a compelling reason to use alpine vs scratch as the base image I could easily be convinced.

stuart-warren:
Sure, so gitlab-ci (opensource integration system) appears to require (no obvious docs that state this) that a container used in a job has a shell :( Each ci runner essentially clones your code, then each job mounts your code into your chosen image and runs any commands you specify. https://docs.gitlab.com/runner/executors/docker.html

The scratch base image doesn't have a shell, so the job just errors out.

I was considering creating 2 separate images but that seemed overkill.

Also considered busybox, but for an extra 3MB you get a package manager etc which gives a little more flexibility if some pre-processing of the file is needed.

Currently image is 10.4MB containing a ~6MB binary

.travis.yml
```diff
@@ -1,8 +1,18 @@
...
+
 go:
   - 1.7
   - 1.8
 install:
   -
 script:
   - ./test
+  - ./docker-build
+after_success:
```
dghubble: You can do this in a better way: https://github.com/coreos/matchbox/blob/master/.travis.yml#L18

dghubble:
Otherwise LGTM

.travis.yml
dghubble: Mostly a nit, doing `"$TRAVIS_BRANCH" == "master"` isn't the worst

stuart-warren: Seems to have used what travis documents https://docs.travis-ci.com/user/docker/#Branch-Based-Registry-Pushes, but the matchbox way looks cleaner

stuart-warren:
rebased and attempted the matchbox way of travis pushing images, not a travis user though

PTAL

sdemos:
looks like everything was successfully built and pushed to quay - https://quay.io/repository/coreos/ct?tab=tags. Thanks for the pr @stuart-warren!

PR #141, opened 2018-02-14
Add support for etcd v3.3

vboginskey:
etcd v3.3.0 was released on 2018-02-01. This PR adds support for it.


vboginskey:
cc: @crawford @dgonyeo @ajeddeloh

ajeddeloh:
Since `gofmt` ends up making diffs nearly unreadable (especially on GH), can you please also mention which fields are added in the commit messages.

ajeddeloh:
It also looks like a lot lot of the fields moved around in terms of order, which also makes the diff harder to read. Can you also ammend it so they don't move? It makes reading the diffs a little easier.

The `ExperimentalCorruptCheckTime` field was added in `08157a9157b0bc095f382b2536c0aac7d50864b3`, was that intended for `86f5e3af860c37f795c2f017101dd80147e90887`?

vboginskey:
Sure, I can make those changes. The diff is easier to read if you append `?w=1` to the GitHub URL: https://github.com/coreos/container-linux-config-transpiler/pull/141/commits/08157a9157b0bc095f382b2536c0aac7d50864b3?w=1.

The `ExperimentalCorruptCheckTime` field is the only [documented](https://github.com/coreos/etcd/blob/v3.3.1/Documentation/op-guide/configuration.md) experimental flag, and I split up the commits based on whether the flags were documented. I can squash the last two commits if that's less confusing.

ajeddeloh:
Oh neat, that `?w=1` change is super useful!

Didn't realize they were document/not documented in that way. That's fine then.

Thanks for the PR!

vboginskey:
Ok, I've removed the sort order changes and listed the added fields in each commit message.

dgonyeo:
Without cross-referencing the etcd docs, this all LGTM

vboginskey:
Sure thing - thanks for the quick review!

PR #144, opened 2018-02-15
doc/dynamic-data: use platforms in dyn. data docs 

ajeddeloh:
The docs builder doesn't have the newest ct version. Change the CLCs in
the docs to just be yaml until the docs builder has ct 0.7.0.


ajeddeloh:
Updated to use the correct platforms. Builder update here: https://github.com/coreos-inc/pages/pull/110

PR #147, opened 2018-03-09
config: emit error when more than one source is set for a file

dgonyeo:
ct should error on a config when a users sets more than one source for a file. Setting no source results in an empty file being created, which seems fine to me.


config/types/files.go
```diff
@@ -101,6 +103,23 @@ func (d Directory) ValidateMode() report.Report {
 	return report.Report{}
 }
 
+func (fc FileContents) Validate() report.Report {
+	i := 0
+	if fc.Remote.Url != "" {
+		i = i + 1
```
ajeddeloh: i++ ?

dgonyeo: Updated

PR #149, opened 2018-03-14
 config/systemd: validate unit and [Intall] section 

euank:
I'm specifically making this PR because I left off the 'install' section on a unit in my config, I included 'enable', and what do you know, the unit didn't do anything.
If I can save a few other people 10 minutes of time with a warning, that seems like a nice thing to do.

This code will miss a case where a unit is enabled and the install section is in a drop-in, but I can't think of any valid reason to do that, so I haven't attempted to not warn in that case.

It is, of course, valid to have units with no "Install" section, but those shouldn't ever be enabled, but rather pulled in via other units wanting or requiring them. I attempted to make the warning read such that it doesn't preclude that alternate possible intention.


config/types/systemd.go
```diff
@@ -56,8 +62,43 @@ func init() {
...
+		for _, section := range parsedUnit {
+			if section.Section == "Install" {
+				noInstall = false
+			}
+		}
+
+		if isEnabled && noInstall {
+			rep.Add(report.Entry{
+				Kind:    report.EntryWarning,
+				Message: fmt.Sprintf("systemd unit %q has no [Install] section; 'enabled' will do nothing", unit.Name),
```
arithx: Should the warning be `'enable(d)' will do nothing` as this triggers if the config has either?

euank: `enable` is deprecated, so I chose to just call it by the supported name

dgonyeo:
Good addition, but a couple things:
- ct uses [Ignition's validation logic](https://github.com/coreos/container-linux-config-transpiler/blob/0c9babbb47e473b3058a942c8f84def6b0824ac3/config/config.go#L56), so if you name the function `Validate` it'll [automatically be called](https://github.com/coreos/container-linux-config-transpiler/blob/0c9babbb47e473b3058a942c8f84def6b0824ac3/config/config.go#L56), and that way when this warning is generated line/column information will be attached.
- This is on a part of the struct shared with Ignition, so if you add this validation logic to Ignition (a `func (Unit u) Validate() report.Report` [here](https://github.com/coreos/ignition/blob/master/config/v2_3_experimental/types/unit.go) would work nicely) then warnings will also pop out from ct because [ct validates the produced Ignition struct](https://github.com/coreos/container-linux-config-transpiler/blob/0c9babbb47e473b3058a942c8f84def6b0824ac3/config/types/converter.go#L61).

Would you mind moving this to the Ignition repo? This PR could still include that test cleanup you did.

euank:
@dgonyeo I'm more wary about adding an error to ignition since backwards compatibility is more of a concern there.
For the warnings, it absolutely makes sense for them to be there, but if I move this to ignition I'd probably also move the error I added here to a warning.

Thanks for the pointers!

dgonyeo:
Unit contents are [already validated](https://github.com/coreos/ignition/blob/bd3876b8f1d627625fe266a0e35256ba05e664e3/config/v2_3_experimental/types/unit.go#L33) in Ignition, so you'd just need to add the warning you included here. Bonus points if you add the validation logic to the 2.2.0, 2.1.0, and 2.0.0 configs as well (it should be mostly copy/pasting).

euank:
This got moved over to the ignition repo; closing over here.

PR #150, opened 2018-03-14
Emit 2.2.0 Ignition configs

dgonyeo:
Updates ct to emit 2.2.0 Ignition configs, and adds the new fields introduced in the 2.2.0 spec.

Shouldn't be merged/released until Ignition v0.22.0 is available in stable CL, but if I remember correctly that'll be the next stable release.


ajeddeloh:
Lets also wait until the refactoring (e.g. https://github.com/coreos/ignition/pull/525 and https://github.com/coreos/ignition/pull/526) emerge first in case they conflict

dgonyeo:
Updated to use the new v0.24.0 Ignition release, so this is now ready for review.

PR #151, opened 2018-04-10
Update README.md with latest version when building from source

WesleyThurner:
Just updated the version in the git clone line, this ensures that people are installing the latest transpiler when following the docs.


WesleyThurner:
I tried to rebase, however additional commits were made.  I am going to close this pull request and create a new one.

PR #154, opened 2018-05-15
config: make getCliArgs handle pointers, make etcd.EnableV2 a pointer

dgonyeo:
Also adds a test to ensure the flag for etcd is set to false.

Fixes https://github.com/coreos/bugs/issues/2428

We should definitely do a full audit of options, I wouldn't be surprised if there are other fields with this issue. Until we do though, this will at least fix the reported issue.

### DIFF
--- a/config/types/common.go
+++ b/config/types/common.go
@@ -113,6 +113,9 @@ func getArgs(format, tagName string, e interface{}) []string {
 				vars = append(vars, getCliArgs(val)...)
 			} else {
 				key := et.Field(i).Tag.Get(tagName)
+				if ev.Field(i).Kind() == reflect.Ptr {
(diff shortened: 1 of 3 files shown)


config/types/common.go
```diff
@@ -113,6 +113,9 @@ func getArgs(format, tagName string, e interface{}) []string {
 				vars = append(vars, getCliArgs(val)...)
 			} else {
 				key := et.Field(i).Tag.Get(tagName)
+				if ev.Field(i).Kind() == reflect.Ptr {
+					val = reflect.Indirect(ev.Field(i)).Interface()
```
ajeddeloh: Why not just `val = *val` ? Or does that not type check?

PR #158, opened 2018-06-05
Support for Compressed, Base64-Enc, Inline Content

akutz:
This patch adds support for local, inline content as compressed, base64 encoded values. For example, the string `Hello, world.` can be compressed with gzip and base64 encoded with the following command:

```shell
$ echo Hello, world. | gzip | base64
H4sIAG/kFlsAA/NIzcnJ11Eozy/KSdHjAgDXu838DgAAAA==
```

A Container Linux Config could use the above contents and write them to the file `/etc/helloworld.txt`:

```yaml
storage:
  files:
    - path: /etc/helloworld.txt
      filesystem: root
      mode: 0644
      contents:
        base64: true
        compression: gzip
        inline: |
          H4sIAB/jFlsAA/NIzcnJ11Eozy/KSdHjAgDXu838DgAAAA==
```

The CT transpiler would transform the above configuration snippet into the JSON below:

```json
{
  "filesystem": "root",
  "path": "/etc/srv/kubernetes/pki/apiserver.key",
  "contents": {
      "compression": "gzip",
      "source": "data:;base64,H4sIAB/jFlsAA/NIzcnJ11Eozy/KSdHjAgDXu838DgAAAA==",
      "verification": {}
  },
  "mode": 420
},
```

Please note the value of the `source` field -- it uses the data URL [scheme](https://tools.ietf.org/html/rfc2397) to indicate the content is base64 encoded. The `compression` field indicates the content is also compressed using gzip. In fact, the following command may be used to verify the data matches the original `Hello, world.` string:

```shell
$ echo H4sIAB/jFlsAA/NIzcnJ11Eozy/KSdHjAgDXu838DgAAAA== | \
  base64 -D | \
  gzip -d
Hello, world.
```


config/types/files.go
```diff
@@ -63,9 +63,11 @@ type File struct {
 }
 
 type FileContents struct {
-	Remote Remote `yaml:"remote"`
-	Inline string `yaml:"inline"`
-	Local  string `yaml:"local"`
+	Base64      *bool  `yaml:"base64"`
```
crawford: I would rather we leave this up to the YAML parser to interpret this via type tagging. See http://yaml.org/type/binary.html for the binary tag.

config/types/files.go
```diff
@@ -63,9 +63,11 @@ type File struct {
...
+	Compression string `yaml:"compression"`
```
ajeddeloh: I'm not a fan of having both `compression` and `remote.compression`. Supporting compression for local/inline files is good, but it should be consistent.

akutz: Hi @ajeddeloh,

That's fine. I just did this to maintain backwards compatibility.

config/types/files.go
akutz: Hi @crawford,

Are you suggesting that I modify `Inline` to be a different type then? Or are you suggesting this as a CT example:

```yaml
storage:
  files:
    - path: /etc/helloworld.txt
      filesystem: root
      mode: 0644
      contents:
        compression: gzip
        inline: !!binary "\
          H4sIAB/jFlsAA/NIzcnJ11Eozy/KSdHjAgDXu838DgAAAA=="
```

config/types/files.go
akutz: Hi @ajeddeloh,

FWIW, I believe if compression is added to local content, it should be removed as a field from remote. As I said, I did it the way I did it to maintain back-compat.

config/types/files.go
crawford: I'm suggesting the CLC example be changed to what you show:

```yaml
storage:
  files:
    - path: /etc/helloworld.txt
      filesystem: root
      mode: 0644
      contents:
        compression: gzip
        inline: !binary |-
          H4sIAB/jFlsAA/NIzcnJ11Eozy/KSdHjAgDXu838DgAAAA==
```

akutz: Okay. I do not know to detect that tag in Go. I will look into it.

ajeddeloh: The yaml parser _should_ handle it automatically

akutz: Hi @ajeddeloh,

Okay, but I still am not sure how I then detect that the string is base-64 encoded and thus set the data URL's `base64` attribute from that? Because *that* is the reason for the boolean in the first place. Without setting that, Ignition's URL parser won't know to base-64 decode the data string.

Ignition uses the Go package [`github.com/vincent-petithory/dataurl`](https://github.com/vincent-petithory/dataurl) to parse URLs for a possible data element. This package properly [detects](https://github.com/vincent-petithory/dataurl/blob/master/dataurl.go#L143-L151) the base64 encoding where Go's doesn't. Unless this is set, the package assumes the data value is URL-encoded. 

Do you see the change I made to the way the value is encoded if the `Base64` field is set to `true`?

```go
if file.Contents.Base64 != nil && *file.Contents.Base64 {
    newFile.Contents = ignTypes.FileContents{
        Source: (&url.URL{
            Scheme: "data",
            Opaque: ";base64," + file.Contents.Inline,
        }).String(),
    }
}
```

The `Opaque` field adds that `;base64,` section and the contents are no longer URL encoded.

crawford: When you tag the type as "binary", the YAML parser will do the conversion for you. By the time the `FileContents` structure is populated, the contents will already be decoded. This mechanism is separate from the dataurl's encoding (e.g. `data://;base64,<data>`). The resulting dataurl won't be base64 encoded, but will instead be URL encoded.

In the future, we could change the dataurl logic to base64 encode contents under certain circumstances (i.e. when the URL encoded size is larger than base64 encoding).

akutz: Hi @crawford,

I see what you're saying. Hmm. Interesting.

akutz: Hi @crawford,

I take that back, I don't see. Are you saying that the gzip'd data would be URL encoded? Is the gzip'd data already base64 encoded? If so, what's the point of URL encoding anything? I must be missing something. I do not see how the binary notation works with the idea of supplying compressed, gzip'd data as local contents.

FWIW, I *am* aware of how it works currently with the data URL per the way `contents.local` and `--files-dir` works.

I suppose what *I'm* trying to supply is a way to provide gzip'd, base64-encoded data that Ignition will know how to decode and decompress before writing to a file.

config/types/files.go
akutz: Hi @ajeddeloh,

Thoughts on this? I don't mind moving the `Compression` field up a level in the struct hierarchy. I just didn't know if y'all were okay with breaking backwards compatibility. Obviously `Compression` shouldn't exist as a field in `Remote` if the former applies to local content as well.

config/types/files.go
crawford: Given the following config:

```yaml
storage:
  files:
    - path: /etc/helloworld.txt
      filesystem: root
      mode: 0644
      contents:
        compression: gzip
        inline: !binary |-
          H4sIAB/jFlsAA/NIzcnJ11Eozy/KSdHjAgDXu838DgAAAA==
```

The YAML library would parse this into the following Go struct:

```go
FileContents {
  Compression: "gzip",
  Inline: "\x1f\x8b\x08\x00\x1f\xe3\x16\x5b\x00\x03\xf3\x48\xcd\xc9\xc9\xd7\x51\x28\xcf\x2f\xca\x49\xd1\xe3\x02\x00\xd7\xbb\xcd\xfc\x0e",
}
```

The inline string is the gzipped string: Hello, world.

CT would then emit the following JSON:

```json
{
  "ignition": {
    "config": {},
    "security": {
      "tls": {}
    },
    "timeouts": {},
    "version": "2.2.0"
  },
  "networkd": {},
  "passwd": {},
  "storage": {
    "files": [
      {
        "filesystem": "root",
        "path": "/etc/helloworld.txt",
        "contents": {
          "source": "data:,H4sIAB%2FjFlsAA%2FNIzcnJ11Eozy%2FKSdHjAgDXu838DgAAAA%3D%3D",
          "compression": "gzip",
          "verification": {}
        },
        "mode": 420
      }
    ]
  },
  "systemd": {}
}
```

Ignition would (with some modification) decompress the dataurl, revealing "Hello, world.".

akutz: Hi @crawford,

Now I actually do see. Thank you. I usually get it. But it also usually takes me a while. My mama says I'm smart in a special way. :)

Okay, so, sounds like I can just get rid of the `Base64` field and how I refactored the encoding logic. I just need @ajeddeloh to respond re: the `Compression` field. @crawford, do you have a preference? Please note I only maintained the previous field to maintain backwards-compatibility.

config/types/files.go
crawford: There is no guarantee of backward compatibility with CT, so we can (and I think we should) move the field. The reason we don't need to worry about backward compatibility is because it's expected that users pin to a specific version of CT for their config. As they upgrade CT over time, they'll need to potentially update their configs. This would be one of those cases.

crawford:
Can you also update the documentation to reflect this new syntax?

akutz:
Hi @crawford,

I'm more than happy to update the docs if you will let me know which docs you'd like me to update.

crawford:
Just this one: https://github.com/coreos/container-linux-config-transpiler/blob/master/doc/configuration.md

akutz:
Hi @crawford,

Done and done. Thanks!

crawford:
One last request: Can you update the commit message and description to reflect that the only change is to allow compression on all file types (not, just remote). I'd leave the example in the body of the commit description (I love verbose history). Maybe something like `config/types: allow compression on files from all sources`.

config/types/files.go
ajeddeloh: Uh, we're not 100% sure that's the case. We failed to decide how to version ct configs and definitely didn't communicate anything along the lines of "hey, this spec is subject to change". I'm not sure what the best path forward is, but I'd be inclined to deprecrate the old field while adding the new one and removing after a set amount of time instead of just removing it with no warning.

akutz:
Hi @crawford,

Done and done. I also updated the failing test. I didn't run the tests locally (dumb, I know), I just compiled it.

akutz:
Hi @crawford,

I also like verbose commit messages. Please see the guidelines I wrote regarding commit messages for one of my projects - http://rexray.readthedocs.io/en/latest/dev-guide/project-guidelines/#commit-messages.

config/types/files.go
crawford: CT has not hit 1.0 and since the spec isn't versioned, it's also subject to the same versioning rules. We can use this opportunity to explain to users how to properly use the tool. I think this will be a good first test since there are probably very few people making use of compression.

akutz:
Hi @crawford,

I just rebased, but one test is still failing. I'm not sure how to fix this as the error appears generated.

```shell
$ make test
Checking gofix...
Checking gofmt...
Checking govet...
Running tests...
--- FAIL: TestParseAndConvert (0.00s)
        Error Trace:    config_test.go:2088
	Error:  	Not equal: report.Report{Entries:[]report.Entry{report.Entry{Kind:0, Message:"path not absolute", Line:4, Column:13, Highlight:""}, report.Entry{Kind:0, Message:"invalid url scheme", Line:17, Column:16, Highlight:""}}} (expected)
			        != report.Report{Entries:[]report.Entry{report.Entry{Kind:0, Message:"path not absolute", Line:4, Column:13, Highlight:""}, report.Entry{Kind:0, Message:"invalid url scheme", Line:18, Column:16, Highlight:""}}} (actual)
			
			Diff:
			--- Expected
			+++ Actual
			@@ -2,3 +2,3 @@
			 path not absolute
			-error at line 17, column 16
			+error at line 18, column 16
			 invalid url scheme
	Messages:	#4: bad report
		
FAIL
coverage: 72.7% of statements
FAIL	github.com/coreos/container-linux-config-transpiler/config	0.018s
?   	github.com/coreos/container-linux-config-transpiler/config/astyaml	[no test files]
?   	github.com/coreos/container-linux-config-transpiler/config/platform	[no test files]
ok  	github.com/coreos/container-linux-config-transpiler/config/templating	0.013s	coverage: 100.0% of statements
ok  	github.com/coreos/container-linux-config-transpiler/config/types	0.010s	coverage: 6.5% of statements
ok  	github.com/coreos/container-linux-config-transpiler/config/types/util	0.011s	coverage: 100.0% of statements
?   	github.com/coreos/container-linux-config-transpiler/internal	[no test files]
?   	github.com/coreos/container-linux-config-transpiler/internal/util	[no test files]
?   	github.com/coreos/container-linux-config-transpiler/internal/util/tools	[no test files]
?   	github.com/coreos/container-linux-config-transpiler/internal/version	[no test files]
make: *** [Makefile:30: test] Error 1
```

akutz:
Hi @crawford,

Nevermind, I figured it out. Now I'm seeing the following error:

```shell
$ make test
Checking gofix...
Checking gofmt...
Checking govet...
Running tests...
ok  	github.com/coreos/container-linux-config-transpiler/config	0.017s	coverage: 72.7% of statements
?   	github.com/coreos/container-linux-config-transpiler/config/astyaml	[no test files]
?   	github.com/coreos/container-linux-config-transpiler/config/platform	[no test files]
ok  	github.com/coreos/container-linux-config-transpiler/config/templating	(cached)	coverage: 100.0% of statements
ok  	github.com/coreos/container-linux-config-transpiler/config/types	(cached)	coverage: 6.5% of statements
ok  	github.com/coreos/container-linux-config-transpiler/config/types/util	(cached)	coverage: 100.0% of statements
?   	github.com/coreos/container-linux-config-transpiler/internal	[no test files]
?   	github.com/coreos/container-linux-config-transpiler/internal/util	[no test files]
?   	github.com/coreos/container-linux-config-transpiler/internal/util/tools	[no test files]
?   	github.com/coreos/container-linux-config-transpiler/internal/version	[no test files]
Checking documentation...
failed while validating docs: non-empty parsing report in examples.md: warning at line 8, column 11
Config has unrecognized key: compressionexit status 1
make: *** [Makefile:30: test] Error 1
```

akutz:
Hi @crawford,

Figured this one out too.

crawford:
@bgilbert I'll let you make the final call on this one. The only point of contention is that this breaks backward compatibility. I'm arguing that we should do it anyway because CT (nor its non-existent spec) is still pre-1.0 and because I don't think many people will be impacted (but have no number to back my claim). I think we should probably explain in our docs that users are expected to pick a version of CT and stick with it since there is no benefit to staying on the leading edge.

lucab:
Jumping a bit late into this discussion, I still find @ajeddeloh's [comment](https://github.com/coreos/container-linux-config-transpiler/pull/158#discussion_r193836706) reasonable.

Perhaps this is a good point to introduce a `kind: cl-config-v1` field, freeze the current schema and proceed with the new/breaking changes in `cl-config-v2-experimental`?

crawford:
@lucab Is the intention to yell at the user if the version doesn't exactly match or to accept multiple versions?

akutz:
Hi @crawford,

May I suggest we merge it as backwards-compatible and decide with a follow-up PR about whether or not to deprecate the old field? That way the feature is added, and the decision regarding how to handle the old field is handled separately.

bgilbert:
@crawford:
> I think we should probably explain in our docs that users are expected to pick a version of CT and stick with it since there is no benefit to staying on the leading edge.

CT is being vendored into things like Matchbox.  Users presumably expect to upgrade Matchbox without breaking their configs, and they don't control what CT version is vendored in.  We're going to have to add proper versioning sooner or later, and I'd rather not initiate that discussion by breaking compat.

So let's do this in a backward-compatible way for now.

crawford:
> CT is being vendored into things like Matchbox. Users presumably expect to upgrade Matchbox without breaking their configs, and they don't control what CT version is vendored in.

It's Matchbox's responsibility to handle versioning though, since it's consuming ct as a library. If we were to add versioning to the ct spec, I would still assume Matchbox would be the one responsible for carrying multiple versions of the library and using the correct one. After all, Matchbox needs to be the one to decide just how many different config versions it's going to support.

> So let's do this in a backward-compatible way for now.

Okay.

bgilbert:
> It's Matchbox's responsibility to handle versioning though, since it's consuming ct as a library.

ct is responsible for parsing CLCs, which makes ct responsible for versioning them. It's unreasonable to require every user of the ct library to vendor N versions of ct (bugs and all) and iterate over them in order to parse a config. We need to take responsibility for the full lifecycle of the CLC spec and not try to export that problem to someone else.

dgonyeo:
The Ignition config package is also used as a library in some places (matchbox also imports this), and afaik there isn't an issue with the fact that Ignition will do the internal translation for library consumers.

I'm in favor of versioning CLCs, I don't think that's complexity that we should be pushing up a layer. After the recent refactor of this in Ignition, I also think that the way translation is modelled over there would be a reasonable pattern to follow here.

config/types/files.go
crawford: @figo Please open an issue if you are having trouble.

figo: @crawford i am happy to do that if you could tell me where should the issue goes, it is ignition, could https://github.com/coreos/terraform-provider-ct/issues be a proper place? thanks

crawford: @figo https://issues.coreos.com

figo: thanks, issue filed, https://github.com/coreos/bugs/issues/2458

ajeddeloh:
The question of whether and how to version CLCs has been an ongoing discussion topic for a while now. Sorry your PR got caught up in that.

If you change the PR to add the new field and not remove the old one and add a validator adding a deprecation notice to the old field, this will LGTM.

I've opened [a new bug](https://github.com/coreos/bugs/issues/2461) to track the CT versioning discussion seperately from this PR.

PR #159, opened 2018-07-10
Fix helloworld oneshot config in example

h8liu:


lucab:
LGTM, thanks for fixing it!

PR #160, opened 2018-07-11
types/{files/dirs/links}: default to root fs

ajeddeloh:
Default to root if no filesystem is specified. It's what you want 99% of
the time.

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -1181,10 +1181,9 @@ func TestConvert(t *testing.T) {
 					},
 					Files: []types.File{
 						{
-							Filesystem: "filesystem1",
-							Path:       "/opt/file1",
-							User:       &types.FileUser{Id: util.IntToPtr(500)},
-							Group:      &types.FileGroup{Id: util.IntToPtr(501)},
+							Path:  "/opt/file1",
+							User:  &types.FileUser{Id: util.IntToPtr(500)},
+							Group: &types.FileGroup{Id: util.IntToPtr(501)},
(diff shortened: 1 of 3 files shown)


ajeddeloh:
@lucab the files section is pretty much the same for clc's and Ignition configs, so its using the Ignition config validation that happens after translation. Since we're filling in the field, it passes that validation.

lucab:
@ajeddeloh I'm not sure if I misunderstood your reply, but I was talking about CLC parsing/validation only (I'm not concerned by Ignition validation here).

ajeddeloh:
For part where the CLC follows the Ignition spec but just in yaml we don't have any validation for the CLC since it would be just duplicating code; we rely on the Ignitoin validation instead. The whole process looks like:

```
CLC (parse)
Unvalidated CLC struct (validate)
Validated CLC struct (translate)
Unvalidated Ign struct (validate)
Validated Ign struct (marshal)
Ignition config
```
And errors can be generated / reported any step of the way. The valdiation that the `storage.files.filesystem` is non-empty normally happens when the Ignition struct is validated.

Does that answer your question?

lucab:
> The validation that the storage.files.filesystem is non-empty normally happens when the Ignition struct is validated

Thanks, I was expecting an higher level validation too. But it looks like we don't do that on purpose (to avoid duplicating dumb logic?). Still, do we want to cover this new empty-replacement with a simple test?

ajeddeloh:
Pushed a test.

PR #161, opened 2018-10-03
types/config: add version field

ajeddeloh:
Add version field. Only allow 0 or unspecified. This starts versioning
the spec in a backwards compatible way.

### DIFF
--- a/config/types/config.go
+++ b/config/types/config.go
@@ -15,6 +15,7 @@
 package types
 
 import (
+	"errors"
 	"net/url"
 
 	ignTypes "github.com/coreos/ignition/config/v2_2/types"
@@ -22,7 +23,14 @@ import (
 	"github.com/coreos/ignition/config/validate/report"
 )
 
+var (
+	ErrInvalidVersion = errors.New("Invalid version. Only version 0 is supported")
+)
+
 type Config struct {
+	// Version is intentionally undocumented. It is only for if we need to make a breaking
+	// change in the future.
+	Version   int        `yaml:"version"`
 	Ignition  Ignition   `yaml:"ignition"`
 	Storage   Storage    `yaml:"storage"`
 	Systemd   Systemd    `yaml:"systemd"`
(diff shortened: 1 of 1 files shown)


config/types/config.go
```diff
@@ -15,14 +15,20 @@
...
 	"github.com/coreos/ignition/config/validate/astnode"
 	"github.com/coreos/ignition/config/validate/report"
 )
 
+var (
+	ErrInvalidVersion = errors.New("Invalid version. Only version 0 is supported")
+)
+
 type Config struct {
+	Version   int        `yaml:"version"`
```
bgilbert: Is there any reason we'd want this to be a semver?

Probably worth noting in a comment that `version` is intentionally undocumented unless and until we need a version `1`.

ajeddeloh: I don't think a semver is necessary and it muddles the json parsing a little. Given we don't plan to use it unless we make breaking changes, any minor (i.e. backwards compatible) changes wont need to bump it. Using semver would imply that we should be bumping it.

I'll add a comment though.

lucab:
@ajeddeloh sorry for the delay, I was mostly offline yesterday. Two general comments on this:
 * we may want to split unmarshaling `version` as its own earlier step, as the `Config` structure itself may change in breaking ways. This can be done now or at the first breaking change at some point in the future.
 * I agree on the non-semver comment. However I still think we should have a kind-string here (as opposed to a version-integer) and default the missing/empty case to `cl-config-v0`, so that we can accommodate any additional flavors (rhcos, fcos, cl, and version bumps) if we need to.

ajeddeloh:
On the first point: Since Ignition spec 3.0.0 won't ship to CL we don't _anticipate_ breaking changes and needing to use this. We'll split that out if that time comes.

On the second point: We're planning on leaving it unversioned unless there is breaking change, so non-breaking changes wont need version bumps. Since FCOS is a different distro, it will have it's own config and we can revisit versioning there.

lucab:
Ack. Note to self: we plan to have the CL / FCOS selector external of the config object itself (e.g. a different binary) instead of embedding the kind in here.

PR #162, opened 2018-10-03
config/types/files: dont' warn on no mode w/append

ajeddeloh:
Do not warn if there is no mode and you are appending.

### DIFF
--- a/config/types/files.go
+++ b/config/types/files.go
@@ -94,7 +94,7 @@ type Link struct {
 }
 
 func (f File) ValidateMode() report.Report {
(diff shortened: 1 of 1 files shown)


bgilbert:
FWIW, when appending to a missing file, Ignition creates the file with the specified mode.

ajeddeloh:
Hrmm. I forgot about that. Kind of a "can't win" situation. I think I might revert and make it an `info` instead? I don't think it deserves a warning since its main use case is for when the file already exists.

PR #165, opened 2018-12-04
Bump Ignition to 0.29.0

ajeddeloh:


config/types/disks.go
```diff
@@ -18,13 +18,13 @@ import (
...
 	"github.com/alecthomas/units"
-	ignTypes "github.com/coreos/ignition/config/v2_2/types"
+	ignTypes "github.com/coreos/ignition/config/v2_3/types"
 	"github.com/coreos/ignition/config/validate/astnode"
 	"github.com/coreos/ignition/config/validate/report"
 )
 
 const (
-	BYTES_PER_SECTOR = 512
+	MEGABYTE = 1024 * 1024
```
lucab: This one too is a `MEBIBYTE`.

config/types/disks.go
```diff
@@ -43,12 +43,12 @@ type Disk struct {
...
-	Label    string `yaml:"label"`
-	Number   int    `yaml:"number"`
-	Size     string `yaml:"size"`
-	Start    string `yaml:"start"`
-	GUID     string `yaml:"guid"`
-	TypeGUID string `yaml:"type_guid"`
+	Label    *string `yaml:"label"`
+	Number   int     `yaml:"number"`
+	Size     string  `yaml:"size"`
+	Start    string  `yaml:"start"`
```
lucab: This and the one above now have a different semantics, what do we do?

ajeddeloh: Ooops forgot the docs change. This is a backwards compatible change* so it shouldn't be a problem

* except for configs with partitions < 1M or not aligned to MBs.

lucab: But the one above (size) now gives a different a different result (sectors vs MiB), isn't it?

ajeddeloh: But they should be equivalent. I think it's fine if a newer version of `ct` emits an equivalent but different config. Otherwise we'd have to use deprecated fields.

ajeddeloh:
Marking as DNM until Ignition with spec 2.3.0 hits CL stable

config/types/disks.go
lucab: You are right, I'm sorry my brain missed the conversion logic in both old and new code.

lucab:
LGTM, feel free to merge whenever unblocked.

dghubble:
Is container-linux-config-transpiler generating Ignition v2.3.0 still planned or no?

ajeddeloh:
We can, though we're pretty swamped with FCOS work at the moment. Is there a pressing need for it?

dghubble:
No pressing need. Mainly just looping for clues if this repo will continue to be the place for YAML->Ignition transforms (for v2 and v3 spec) or if that's not the goal. I can take this to some discussion forum so as not to derail.

ajeddeloh:
The Fedora CoreOS equivalent will be in another repo. Currently it's at https://github.com/ajeddeloh/fcct but it'll move over to the coreos github org once it's ready. That's where all the spec v3 stuff is happening and the spec v2 stuff will remain here.

lucab:
Self-note: Ignition >= 0.29.0 reached ContainerLinux stable channel with release 2023.4.0, this is safe to merge now.

PR #167, opened 2019-09-10
config/types/etcd: support 3.4

JTarasovic:
this change adds an `Etcd3_4` config so that `ct` can support etcd 3.4.

it could definitely do with a second set of eyes making sure I've captured the changes appropriately. here's the diff between the 2 structs: https://gist.githubusercontent.com/JTarasovic/a3e9db802b9aead6af69abf274c75c82/raw/750710a37f756a5fe9be2882c2733ef14c1d6ead/diff.txt


JTarasovic:
coreos/bugs#2612

PR #168, opened 2020-03-24
config/types: Fix Convert to output v2.3.0 spec

dghubble:
* Fix Convert to output v2.3.0 spec and fix tests
* Issue: Ignition Config types were updated to v2.3.0 in #165, but still convert a Container Linux Config to v2.2.0 Ignition. It seems, the Convert version and tests were missed, compared with previous spec updates

Related: https://github.com/coreos/container-linux-config-transpiler/commit/d9c733ca08859c15fef5e96c89d23d1c31d86037#diff-8f256984855e8444054b0ecb38776f8e

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -893,7 +893,7 @@ func TestConvert(t *testing.T) {
 	}{
 		{
 			in:  in{cfg: types.Config{}},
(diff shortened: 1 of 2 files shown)


dghubble:
Probably @lucab would have context

lucab:
Oh yes, #165 missed the converter bump and I missed that when reviewing. Good catch, LGTM.

lucab (approved):

